PR #50, opened 2024-03-13
Add Elastic environment variables to README

elvbom:
Elastic variables have been added to the README since they are needed in the .env file.


irony (commented):
Bra!

PR #83, opened 2024-04-25
Feat/nlm ingestor

irony:
This is using the nlm-ingestor docker file results to extract markdown. Work in progress


.gitignore
```diff
@@ -1,12 +1,13 @@
 node_modules
 dist
+.env
+.aider*
+prisma/
```
Greenheart, line 5: This is probably wrong. We want to keep `prisma/migrations/*`, `prisma/schema.prisma` and more in version control.

`prisma/generated` could be gitignored, but the rest should be kept in git.

output/table-15-CO2 (kTons) 300.png
Greenheart, line 1: Should `output/*` be gitignored?

src/lib/DiscordWorker.ts
```diff
@@ -99,5 +100,7 @@ export class DiscordWorker<T extends DiscordJob> extends Worker<any> {
       concurrency: 10,
       ...options,
     })
+
+    this.queue = new Queue(name, { connection: redis })
```
Greenheart, line 104: Since we have multiple instance of each worker, this would also create multiple queues with the same name. I was wondering if this would be a problem, but it seems like Bull MQ handles this.

From their documentation: https://docs.bullmq.io/guide/queues
> When you instantiate a Queue, BullMQ will just upsert a small "meta-key", so if the queue existed before it will just pick it up and you can continue adding jobs to it.

irony: No the worker object is global the job instance is many. It should be fine to have the queue on the worker.

PR #92, opened 2024-04-29
upgrade openai package to 4.38.5

elvbom:


Greenheart (approved):
The `openai` package seems to follow semver, based on the changelog.

This version update might allow us to experiment with the new model `gpt-4-turbo-2024-04-09` which seems to be one of the first "stable" releases.

https://platform.openai.com/docs/models/gpt-4-turbo-and-gpt-4

PR #99, opened 2024-04-30
Update gpt model from gpt-4-1106-preview to gpt-4-turbo

elvbom:


src/workers/extractEmissions.ts
```diff
@@ -36,7 +36,7 @@ const worker = new Worker(
         { role: 'system', content: prompt },
         { role: 'user', content: pdfParagraphs.join('\n\n') },
       ],
-      model: 'gpt-4-1106-preview',
+      model: 'gpt-4-turbo',
```
Greenheart, line 39: Do we want to automatically get bumped to the new version when it becomes available?

If we specify `gpt-4-turbo-2024-04-09`, we will get to keep using the same model up to 3 months after a new one is released. Perhaps this would make it more stable, and explicit when a model upgrade happens?

List of models:
https://platform.openai.com/docs/models/gpt-4-turbo-and-gpt-4

More on OpenAI continuous model upgrades:
https://platform.openai.com/docs/models/continuous-model-upgrades

irony: I guess the newer models will be better so I think we are fine with getting new ones when they arrive.

irony (approved):
Bra! Ska bli kul att testa. Jag mergar direkt.

PR #100, opened 2024-04-30
add deferReply to /parse and /pdfs

elvbom:
Add deferReply to the slash commands /parse and /pdfs to prevent Discord from showing the "The application did not respond" error when using said commands


src/discord/commands/pdfs.ts
```diff
@@ -11,11 +11,13 @@ export default {
         .setRequired(true)
     )
     .setDescription(
-      'Skicka in en årsredovisning och få tillbaka utsläppsdata.'
+      'Skicka in flera årsredovisningar och få tillbaka utsläppsdata.'
```
Greenheart, line 14: If we remove the `/pdf` command, this should probably say `en eller flera årsredovisningar`

src/discord/commands/parse.ts
```diff
@@ -12,7 +12,9 @@ export default {
     ),
 
   async execute(interaction) {
-    console.log('pdf')
+    console.log('parse')
+    await interaction.deferReply({ ephemeral: true })
```
Greenheart, line 16: Very nice find! I haven't tested this locally yet (still setting up my environment), but it looks good and makes sense based on what I can find in the `discord.js` API docs

https://discord.js.org/docs/packages/discord.js/main/CommandInteraction:Class#deferReply

Greenheart (changes requested):

src/discord/commands/pdfs.ts
elvbom: True. Since it's a separate PR that hasn't been merged yet, I think that would be a better place for the change though.

Greenheart, line 14:
```suggestion
      'Skicka in en eller flera årsredovisningar och få tillbaka utsläppsdata.'
```

irony:
Sorry jag lyckades paja den här PR:n med mina ändringar. Måste sluta pusha till main.

Greenheart:
No worries - it was easily resolved! I know the feeling when working alone (if only for a day) and you want to see the results quickly :)

PR #101, opened 2024-04-30
remove /pdf command

elvbom:
Remove /pdf since /pdfs covers its functionality and /pdf contains bugs


Greenheart (approved):
I think it makes sense to have one way to do things. As long as the `/pdfs` command also can handle only a single report (or an array of one item), I'm happy :)

irony:
Approved!

PR #102, opened 2024-04-30
Ensure previous job was successful and returned actual data

Greenheart:
This PR solves `TypeError: Cannot read properties of null (reading 'length')` which was triggered in at least one instance at http://localhost:3000/admin/queues/queue/pdf2Markdown/93?status=failed

It seems like `completed` jobs can still get `null` as their return value. It might be a symptom of previous jobs not finishing as expected.

Looking at the error from above, the input data looks odd. According to the type definition at https://github.com/Klimatbyran/garbo/blob/543c3bf54e7aeb6500f825fdd131b7063f1daf61/src/workers/pdf2Markdown.ts#L104, `existingId` should always be a string, but as you can see in the screenshot below, it's `undefined`.

image

This is what caused the original error.

### DIFF
--- a/src/workers/pdf2Markdown.ts
+++ b/src/workers/pdf2Markdown.ts
@@ -4,7 +4,6 @@ import { pdf2Markdown, splitText } from '../queues'
 import elastic from '../elastic'
 import llama from '../config/llama'
 import discord from '../discord'
-import { TextChannel } from 'discord.js'
 
 const minutes = 60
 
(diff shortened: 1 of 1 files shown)


Greenheart (commented):
I think I found another potential problem with https://github.com/Klimatbyran/garbo/pull/102/files#diff-03cc518e3d991a951973b23ab5af7239fcd6bdbf066f54f360b92db5c5b7d2b8L122-L129

What if `previousJob` results in undefined with the new logic, while we simultaneously get `existingId` (like the type definition says we should expect).

This might cause the same error again, attempting to read `text.length` even though it's still `null` from the initialization on https://github.com/Klimatbyran/garbo/pull/102/files#diff-03cc518e3d991a951973b23ab5af7239fcd6bdbf066f54f360b92db5c5b7d2b8R118.

Potential solution: Maybe we could force-retry the parsing job? Thus modifying the condition `} else if (!existingId) {` to perhaps `else if (!previousJob || !existingId) {`

irony:
Apart from the question above the PR is approved

irony:
Jag mergade denna nu efter att ha dubbelkollat att pdfHash inte är en asynkron funktion

Greenheart:
Ah! Yeah, I found out in the editor and verified with the implementation of `pdfHash()`, but I should have written that explicitly to make it clear why that was included in the same PR. This could have been an error too :)

PR #103, opened 2024-05-02
fix: only handle commands and interactions from one instance

irony:
This PR aims to solve the problem when multiple workers acknowledge the interaction at the same time resulting in worker restart. This will hopefully solve some of the instability issues we have seen where some of the jobs are stalling.

image

### DIFF
--- a/src/config/discord.ts
+++ b/src/config/discord.ts
@@ -2,5 +2,6 @@ export default {
   token: process.env.DISCORD_TOKEN,
   clientId: process.env.DISCORD_APPLICATION_ID,
   guildId: process.env.DISCORD_SERVER_ID,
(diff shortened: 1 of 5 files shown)


src/config/discord.ts
```diff
@@ -2,5 +2,6 @@ export default {
   token: process.env.DISCORD_TOKEN,
   clientId: process.env.DISCORD_APPLICATION_ID,
   guildId: process.env.DISCORD_SERVER_ID,
+  worker: process.env.DISCORD_WORKER === 'true',
```
Greenheart, line 5: Seems like a new env variable. Would be good to also add to the README!

Also, is this an optional variable? When is that useful? Maybe dev vs prod?

Greenheart (changes requested):

PR #105, opened 2024-05-02
Add new env variable to README

Greenheart:
Resolves https://github.com/Klimatbyran/garbo/pull/103#discussion_r1587552270

`DISCORD_WORKER` seems like an optional env variable. Is this correct? Is this used differently between for example dev and prod?


Greenheart:
Ah, yeah it makes sense from the code. Would likely be worth clarifying the purpose via the env variable name.

Any suggestion?

irony:
It is an internal optional variable used by the infrastructure. I'm not sure we need to clarify its use in the README. Maybe it's better to find a way to not use enviroment variable at all for this purpose.

PR #106, opened 2024-05-02
feat: add facit from csv

irony:
Add facit to response to make the review process easier. Later we can do this automatically.


src/data/facit.csv
```diff
@@ -1,151 +1,151 @@
...
-N           ,Lagercrantz Group                    ,https://www.lagercrantz.com/sites/lagercrantz/files/files/Lagercrantz%20Group%20H%C3%A5llbarhetsrapport%202022_23_1.pdf                                                                                     ,https://www.lagercrantz.com/sites/lagercrantz/files/files/Lagercrantz%20Group%20H%C3%A5llbarhetsrapport%202022_23_1.pdf                                                                                                                                 ,Brutet räkenskapsår                                                                                                                          ,Nej                                                                                                     ,Nej - rapporten läses fortfarande snett     ,
```
Greenheart, line 2: Random detail (maybe not relevant at all):

The same 2023 report link is used for both 2022 and 2023 for AAK: https://www.aak.com/siteassets/sustainability/sustainability-reports/aak-pai-report-2023.pdf

Does this mean the report was not found for 2022?

src/models/companyEmissions.ts
```diff
@@ -2,20 +2,20 @@ export type YearEmissions = {
...
   scope2: {
-    emissions: string
-    unit: number
-    mb: string
-    lb: string
-    baseYear: string
+    emissions: number
+    unit?: number
+    mb: number
+    lb: number
```
Greenheart, line 12:
```suggestion
    /** Location-based */
    lb: number
```

src/models/companyEmissions.ts
```diff
@@ -2,20 +2,20 @@ export type YearEmissions = {
...
   }
   scope2: {
-    emissions: string
-    unit: number
-    mb: string
-    lb: string
-    baseYear: string
+    emissions: number
+    unit?: number
+    mb: number
```
Greenheart, line 11:
```suggestion
    /** Market-based */
    mb: number
```

src/startWorkers.ts
```diff
@@ -1,5 +1,6 @@
 import dotenv from 'dotenv'
 dotenv.config() // keep this line first in file
+process.env.WORKER = 'true'
```
Greenheart, line 3: In this context, I think it makes sense to use process.env to make the discord bots in a certain way.

The env variable makes sense too, since it tells all code it's running in a worker context.

src/lib/facit.ts
```diff
@@ -0,0 +1,77 @@
...
+          csv({
+            mapHeaders: ({ header }) => header.trim(),
+            mapValues: ({ value }) => value.trim(),
+          })
+        )
+        .on('data', (data) => all.push(data))
+        .on('end', () => {
+          const found = all.find((result) => result['URL 2023'] === url)
+          console.log('found', found)
+          resolve(mapFacitToCompanyData(found))
```
Greenheart, line 64: Since we only want one row, it feels a bit wasteful to parse the entire csv file.

Since we get one row back at a time, I guess one solution could be to move the url check straight into the data callback:

```ts
.on('data', (data) => {
  if (data['URL 2023'] === url) {
    // IDEA: Could we abort the readable stream as soon as we find the data we want?
    // For that we might need to get a reference to it, but it must be doable right?

    resolve(mapFacitToCompanyData(found))
  }
})
```

With this solution, we wouldn't need the `all` array, or the `.on('end', () => {})` callback.

A similar idea is if we could read the csv file only once and keep the parsed data in memory? Or is it worth keeping the memory footprint as low as possible? In that case maybe just the solution above is sufficient.

In the future with many thousands of companies, the facit should probably anyway be moved to a DB of some kind to make it easier to query.

Greenheart (approved):
Looking good from what I can tell!

One optional change, but feel free to ignore it or move to a separate issue.

src/lib/facit.ts
irony: This is a solution only for now- as you say- we will probably use a database for this later on. Parsing from memory takes milliseconds so I wouldnt say we need to worry about this right now.

src/data/facit.csv
irony: Good catch, this was handled by @Almendra-ab  now.

PR #115, opened 2024-05-11
Feat/parse multiple pdfs

irony:
Now you can use /parse with multiple urls and it works in similar way as with /pdfs


Greenheart (approved):
Looking good based on a diff review. Haven't tested it, but we can always make improvements in further PRs

PR #119, opened 2024-05-16
Feat/bun

irony:
Switch to bun instead to remove a lot of issues with resolving typescript module,esnext,commonjs problems


Greenheart (approved):
Tested locally and it runs as expected, apart from elasticsearch. As we said in discussions, we will need to switch to a more recent fork like opensearch.

PR #128, opened 2024-06-03
feat: add fiscal year as followup

irony:
Add fiscalYear as property on company so we can accurately report this data to Wikidata

### DIFF
--- a/src/prompts/followUp/fiscalYear.ts
+++ b/src/prompts/followUp/fiscalYear.ts
@@ -1,0 +1,13 @@
+const fiscalYear = `
+Extract the company fiscal year. Sometimes companies have broken fiscal year.  Example: 1 apr -> 31 mar means: startMonth = 4, endMonth = 3. Standard is 1 jan -> 31 dec. 
+
+Example:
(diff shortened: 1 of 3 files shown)


elvbom (approved):

src/prompts/followUp/fiscalYear.ts
```diff
@@ -0,0 +1,13 @@
+const fiscalYear = `
+Extract the company fiscal year. Sometimes companies have broken fiscal year.  Example: 1 apr -> 31 mar means: startMonth = 4, endMonth = 3. Standard is 1 jan -> 31 dec.
```
Greenheart, line 2: Will `->` be parsed correctly as an arrow? Or would it be better to write an explicit `to` instead, replacing `->`?

Greenheart (commented):
Apart from comment, this looks good! :)

src/prompts/followUp/fiscalYear.ts
irony: It should be fine.

PR #129, opened 2024-06-03
Create employees

Almendra-ab:
Extract total number of employees from annual report


src/prompts/followUp/employees
```diff
@@ -0,0 +1,17 @@
+const employees = `
+Extract the total number of employees that the company had in the given period. Add it as field employees. In the report, this can be referred to as Average value of employees. Make sure that it is the total number of employees, not a subset such as one gender or employees in only one country. Extract this data for all available years.
```
Greenheart, line 2: I assume "Average value of employees" is an intentional word choice to match the reports?

Greenheart (approved):
Apart from my question, this looks like a good addition :)

PR #130, opened 2024-06-04
Adding number of employees to json

Almendra-ab:
The prompt asks for data, this is to also reflect this info in the json

### DIFF
--- a/src/prompts/extractJson.ts
+++ b/src/prompts/extractJson.ts
@@ -87,6 +87,13 @@ const prompt = `I have previously sent a text for analysis by GPT-4. The respons
       "unit": "SEK"
     }
   ],
+  "employees": [
+    {
+    "year": "2021", 
+    "value": "12345",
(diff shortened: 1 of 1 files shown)


src/prompts/extractJson.ts
```diff
@@ -87,6 +87,13 @@ const prompt = `I have previously sent a text for analysis by GPT-4. The respons
       "unit": "SEK"
     }
   ],
+  "employees": [
+    {
+    "year": "2021", 
+    "value": "12345",
+    "unit": "FTE"
```
Greenheart, line 94:
```suggestion
      "year": "2021", 
      "value": "12345",
      "unit": "FTE"
```

Probably not important, but always good to use consistent indentation to avoid confusion

Greenheart (approved):
Hopefully FTE (Full time Equivalent) is understood based on the context. Let's test and see if that needs to be clarified.

PR #141, opened 2024-06-13
feat: add cache for api

irony:
5 mins cache to save bandwidth between pages


Greenheart (approved):
Looking good! Haven't used `apicache` before but it seems like a nice library.

PR #142, opened 2024-06-13
Feat/format

irony:


src/prompts/reflect.ts
```diff
@@ -48,30 +48,11 @@ field empty if the data is not verified by Wikidata.
 ONLY WRITE IN SWEDISH! The data will be shown on a swedish site called Klimatkollen.se.
 If the original texts are written in English, translate to Swedish.
 
-**Example**: Always generate this exact JSON structure, never use the data from the example.
+**Example**: Always generate this exact JSON structure but never use the data from the example.
```
Greenheart, line 51: Might be an important distinction - good catch!

Greenheart (approved):
Only reviewing the diff for now. Seems like good improvements both to the code structure and the prompt. Well done! :)

PR #164, opened 2024-06-18
fix: duplicate entries in the api

irony:


Greenheart (approved):
Good idea to remove the pdfHash since it was not used.

PR #174, opened 2024-06-19
Feat/smaller prompts

irony:
Try to reduce the size of the prompts and the resulting objects


src/prompts/reflect.ts
```diff
@@ -29,50 +29,6 @@ The output format will be reviewed once more in later steps so you can include c
 IMPORTANT to verify all fields and compare them to the wikidata object. If they are correct, mark them as verified with the link to the wikidata url.
 This is super important! You will recieve $200 payment for each verified field.
```
olavG91, line 30: Byt recieve till receive.
Tror inte GPT bryr sig särskilt, men den kanske blir mindre förvirrad :)

Greenheart (approved):
Looking good from a quick look. Thanks for working with this!

PR #189, opened 2024-06-28
Fix/update dependencies

### LINKED ISSUE #187: Update dependencies
## Remove unused:

- @google/generative-ai

## Replace npm package with Node.js native replacement:

- `uuid`, which can be replaced with [crypto.randomUUID()](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID) in Node.js >= 19

## Fix security warnings

Greenheart:
Fix #187 

- Remove some unused dependencies
- Update all dependencies to the latest non-major versions, fixing several security warnings
- Replace `uuid` package with native uuid v4 generation available in Node.js >= 19. Both alternatives implement the same RFC, but native Node.js method is faster and doesn't require an additional npm module. Learn more: https://www.nearform.com/insights/new-crypto-capabilities-in-node-js/


irony (approved):
Ser bra ut!

PR #203, opened 2024-08-15
Feature: Setup new company database + import existing data + basic API

irony:


prisma/generated/schemas/aggregateCompany.schema.ts
```diff
@@ -0,0 +1,29 @@
+import { z } from 'zod'
```
Greenheart, line 1: We should likely add `prisma/generated` to .gitignore.

irony: 👌🏻 Add a script to the docker file and ignore the generated folder?

irony: `RUN npx prisma generate` before cmd?

prisma/schema.prisma
```diff
@@ -0,0 +1,182 @@
...
+  sectorCode      String
+  sectorName      String
+  groupCode       String
+  groupName       String
+  industryCode    String
+  industryName    String
+  subIndustryCode String
+  subIndustryName String
+  companies       Company[]
+}
```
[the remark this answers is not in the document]
Greenheart: This is now resolved. We will store the string labels as part of other i18n strings. Right now they are committed to the repository but could be moved to the i18n files in the future.

src/workers/downloadPDF.ts
```diff
@@ -12,6 +12,10 @@ class JobData extends Job {
   }
 }
 
+function hashPdf(pdfBuffer: Buffer): string {
+  return crypto.createHash('sha256').update(pdfBuffer).digest('hex')
+}
+
```
Greenheart, line 18: Do we want to use the same `hashPpf()` implementation for both `downloadPDF` and `pdf2Markdown` to avoid potential mismatch errors when we check the content in the future?

PR #216, opened 2024-09-19
Feat/prisma aider2

irony:


src/lib/prisma.ts
```diff
@@ -32,8 +32,8 @@ export async function updateScope1(
           ...scope1,
           unit: tCO2e,
           metadata: {
-            create: {
-              ...metadata,
+            connect: {
+              id: metadata.id,
```
Greenheart, line 36: Ah, this makes so much sense in hindsight! Thanks for fixning this :)

PR #224, opened 2024-10-09
remove --import flag from start script

elvbom:
When starting Garbo locally, I had to update the start script by removing 'node --import' (due to not being a valid option for node.js). Something wrong on my end or something that should be changed?

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,8 +7,8 @@
     "node": ">=22.0.0"
   },
   "scripts": {
-    "start": "node --import tsx src/index.ts",
-    "workers": "node --import tsx src/startWorkers.ts",
+    "start": "tsx src/index.ts",
+    "workers": "tsx src/startWorkers.ts",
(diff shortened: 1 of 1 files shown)


Greenheart (commented):
I think this is just due to outdated documentation from our side :sweat_smile: 

Which version of Node.js are you using? Our docker image is using `node:alpine` which resolves to Node 22. I'm also using Node 22 locally.

We are using a new CLI option available since Node.js `v18.18.0` to automatically load the [tsx](https://github.com/privatenumber/tsx/) package which allows us to run TS without having to manually transpile it into JS first. Using the `tsx` command directly also works, but we use `node --import tsx` to be able to run Node directly and only use tsx for transpiling TS.

https://nodejs.org/docs/latest/api/cli.html#--importmodule

We should likely add `engines` to package.json to ensure a specific version is used. More details: https://stackoverflow.com/a/29349773

PR #226, opened 2024-10-17
Enabled cors function for the entire API

ZaraBaghdasaryan:
This was needed to make the API work locally.


Greenheart (approved):
The latest commit looks good, but I think need to do a rebase merge to prevent merge conflicts. I'll see what can be done.

PR #269, opened 2024-11-21
feat: Add a wrapper lib around our vector DB to simplify the code and future updates

Greenheart:
This simplifies the worker code interacting with the vector DB.

NOTE: I opted to let `vectorDB.addReport()` accept both `markdown` and `paragraphs` as parameters, rather than just `markdown`. The reason is that it made more sense to log the number of paragraphs in the `indexMarkdown` worker. One more parameter to the `addReports()` method is worthwhile since it makes the code much more readable, logging the number of paragraphs before we attempt to save to the vector DB.


hugo-nl (approved):
## Nice work! 🚀 
🥇  Really nice to have all of this logic in one file
🥇 Better readability and more concise than before
✅ Approved merge to staging at your convenience 🎉

PR #278, opened 2024-11-22
Feat/add query texts to enhance vector search

### LINKED ISSUE #275: Improve vector search - By adding intentional queryTexts
⚙️ Currently we are sending the entire prompt context for the vectorsearch. The prompt is however quite long and contain lots of irrelevant words for our target data.
⚙️ If we try export more precise keywords related to the target value we are looking for ex: scope1 then we might retrieve better result improving the relevant markdown returned from the vector database. This would in second effect improve the context sent to askStream in followUp where we ask the llm to extract the correct value

⚡ If we add this to each prompt file we can easily experiment to find the with the best search configuration

### LINKED ISSUE #277: Startdate cannot be after enddate
🐛 This bugg occurs when a company does not have the standard jan-dec accounting year. 
⚡ Solution is to check if startmonth > endmonth and then start year is --> year -1

hugo-nl:
🌞 This PR introduces intentional "queryTexts" which should be more effective for search vector retrieval than the entire prompt 
👍 Also addresses bugg #277

fixes #275 
fixes #277

### DIFF
--- a/src/lib/reportingPeriodDates.ts
+++ b/src/lib/reportingPeriodDates.ts
@@ -29,7 +29,7 @@ export function getReportingPeriodDates(
   startMonth: number,
   endMonth: number
 ) {
-  const startYear = startMonth > endMonth ? year - 1 : year
+  const startYear = startMonth >= endMonth ? year - 1 : year
   const endYear = year
 
   const start = format(new Date(startYear, startMonth - 1, 1), 'yyyy-MM-dd')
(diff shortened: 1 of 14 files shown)


Greenheart (commented):
This is a good improvement, and it seems to work better than what we had before :)

Some minor changes could be made, but otherwise this is almost ready to merge

Greenheart (approved):

PR #283, opened 2024-11-25
Fix #281 - Only show Discord review button for actual changes that need to be reviewed

### LINKED ISSUE #281: Only show the Discord review button(s) if there are actual changes
For example in the case of the screenshot below, we should simply log that there were no changes found. Then could complete this job.

image

Greenheart:


hugo-nl (approved):
Nice !! ✅ 🎉

PR #284, opened 2024-11-25
Update chunking strategi

### LINKED ISSUE #282: Try alternative chunking strategi
📃 Currently we are chunking per new heading (see indexMarkdown)
📃This might be to big pieces capture granular context

⚡ Lets try larger chunks with a little overlap between them

hugo-nl:
✂️ Previously we divided chunks by "###n" (header sections)
✂️ With this approach we divide the text into chunks of 1000 characters with 200 character overlap

📈 From my experiments the results where substantially better with this method

Ex:

### Result with previous chunking method
before_new_chunking

### Result with new chunking method (3/3 categories identified) 🚀 
after_new_chunking

Tried a few more as well. Here it is 100% https://www.add.life/media/nwepx4r4/addlife-annual-report-2023.pdf

fixes #282

### DIFF
--- a/src/lib/vectordb.ts
+++ b/src/lib/vectordb.ts
@@ -11,9 +11,9 @@ const collection = await client.getOrCreateCollection({
   embeddingFunction: embedder,
 })
 
-async function addReport(url: string, markdown: string, paragraphs: string[]) {
-  const ids = paragraphs.map((p, i) => url + '#' + i)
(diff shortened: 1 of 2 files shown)


Greenheart (commented):
Great find! :tada:

Some minor comments and then we can merge.

Greenheart (approved):

PR #287, opened 2024-11-26
Update chromadb JS client to work with latest docker image. Also update other dependencies.

Greenheart:
Verified locally that the latest versions work together.


hugo-nl (approved):
Infra win! 🔥

PR #289, opened 2024-11-26
Feat/nlm ingestor response parsing

Greenheart:
- Parse output from nlm-ingestor to fix edge cases and make the integration more robust
- Remove unused parsing step since full_row is a type of `TableRow.type`, rather than a `Block.tag`


hugo-nl (approved):
Hero! Super win to get typings for this 🤩

PR #292, opened 2024-11-27
Avoid table screenshot name collisions by using random ID

### LINKED ISSUE #280: Table screenshots are overwritten if multiple tables have the same name
## The problem:
As in the case with the screenshot below, multiple screenshots are found with the same name, which causes only the last one to get saved and used by further steps in the pipeline.

We save two images with the same name: `/tmp/table-32-focus-area__ambition_zero_carbon.png`

## Proposed solution: 
We already add the page number to the file name to make sure we have something unique. However, in this case we need something more.

This is the line that needs to be updated: https://github.com/Klimatbyran/garbo/blob/27bcfcaf97e438382281592cc9a2ccae76586691/src/lib/pdfTools.ts#L122

A related problem could arise if multiple companies have similar table names/captions. This could be fixed if we added the `wikidataId` to separate the files.

image

Greenheart:
Fix #280

### DIFF
--- a/src/lib/pdfTools.ts
+++ b/src/lib/pdfTools.ts
@@ -13,8 +13,6 @@ import {
   ParsedDocumentSchema,
 } from './nlm-ingestor-schema'
 
-const MAX_LENGTH_TABLE_NAME = 50
-
 async function getPngsFromPdfPage(stream: Buffer) {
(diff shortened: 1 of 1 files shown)


irony:
Ser bra ut!

irony (approved):

PR #306, opened 2024-11-27
Fix/diff only changes in section

irony:
This pull request includes several updates to the container images and improvements to the codebase for handling differences between JSON objects and schema validation. The most important changes include updating the container image versions, modifying the JSON diff logic, and adjusting the schema validation.

Improvements to JSON diff logic:

* [`src/workers/saveToAPI.ts`](diffhunk://#diff-0c7c3a577b4209a9dd0b582d26ce031ca4b3977831e8bd6db7a720d4483a1519R40-R69): Modified the `askDiff` function to create `before` and `after` objects with only the relevant fields, and updated the prompt to use "NO_CHANGES" instead of "NO CHANGES".
* [`src/workers/saveToAPI.ts`](diffhunk://#diff-0c7c3a577b4209a9dd0b582d26ce031ca4b3977831e8bd6db7a720d4483a1519L107-R118): Adjusted the condition to check if the diff includes 'NO_CHANGES' instead of being exactly 'NO CHANGES'.

### DIFF
--- a/k8s/api.yaml
+++ b/k8s/api.yaml
@@ -14,7 +14,7 @@ spec:
         app: garbo
     spec:
       containers:
-        - image: ghcr.io/klimatbyran/garbo:3.1.20 # {"$imagepolicy": "flux-system:garbo"}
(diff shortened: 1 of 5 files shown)


Greenheart (approved):
Looking good!

PR #342, opened 2024-11-28
Add dev-reset script

### LINKED ISSUE #264: Add dev scripts for resetting postgres, redis and chroma
This would be useful for local development.

Could be documented in the README.

These should be implemented as two npm scripts, maybe called something like `reset:postgres` and `reset:redis`?

## To reset postgres
We can use the `resetDB()` helper which deletes all data. Useful during development.

```ts
apiRouter.get('/reset-db', async (req, res) => {
  await resetDB()
  res.send({ ok: true })
})
```

## Here's how we can reset the redis DB:

Execute `bash` in the given container and send it a specific command: `redis-cli flushall` which removes all data from redis.

```bash
docker exec -it garbo_redis bash -c "redis-cli flushall"
```

Or maybe an even better option would be to allow clearing specific queues, which could be accomplished via the bullmq Node API

## How to reset chroma:
- Either do a full reset by deleting and re-creating the container (including its volume)
- Or remove just the relevant report URL:s

Greenheart:
Simplify local development when testing reports and resetting especially `chroma` and `redis`

Fix #264

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,8 @@
     "dev": "concurrently \"npm run dev-board\" \"npm run dev-workers\"",
     "import": "node --import=tsx --env-file=.env.development scripts/import-spreadsheet-companies.ts",
     "test": "DOTENV_CONFIG_PATH=./.env.test jest",
-    "prisma": "dotenv -e .env.development -- prisma"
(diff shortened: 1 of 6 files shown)


hugo-nl (approved):
## This is gonna be great for development 🚀 🥇

PR #350, opened 2024-11-28
Feat: Add wait step for base image workflow in CI job

irony:


Greenheart (approved):
Thanks! This will be a great DX improvement :)

PR #359, opened 2024-11-29
Feat/fix-for-reset-dev-script

hugo-nl:
🚅 Noticed the reset for redis did not work for me --> added docker cmd
🚅 Expanded with resetAll func for quick restart

### DIFF
--- a/scripts/dev-reset.ts
+++ b/scripts/dev-reset.ts
@@ -47,7 +47,7 @@ function confirm(message: string): Promise<boolean> {
 async function resetRedis() {
   const commands = [
     'docker exec garbo_redis redis-cli flushall',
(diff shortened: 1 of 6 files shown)


scripts/dev-reset.ts
```diff
@@ -44,6 +44,29 @@ function confirm(message: string): Promise<boolean> {
   )
 }
 
+async function resetRedis() {
+  const commands = [
+    'docker exec garbo_redis redis-cli flushall',
+    'podman exec -it garbo_redis bash -c "redis-cli flushall"',
```
Greenheart, line 50: Good idea!

Here's a simplified command for podman
```suggestion
    'podman exec garbo_redis redis-cli flushall',
```

PR #378, opened 2024-12-03
Only pass parsed json data explicitly when needed, and remove from other jobs

Greenheart:
This should reduce reduce the size of the job data, and reduce storage and memory required for the queue.

Follow up for https://github.com/Klimatbyran/garbo/pull/302 - it seems like that change got lost somewhere when merging staging and main.


hugo-nl (approved):
Nice!

PR #380, opened 2024-12-03
Diff similar structure to allow saving `economy` data even if nothing has been saved before

Greenheart:
This fixes the two diffing-bugs - 1) not possible to save `economy` data and 2) inaccurate and unreliable diffs for `emissions` - introduced last week. However, the code should probably be cleaned up a bit.

I'm considering transforming the `[before, after]` data in the beginning of the worker, so we can make the diff against the same data as we later save. This would also have the benefit of simplifying the process of saving each data point, since we could refactor this worker to only do the structural data transformation to match the API expectations in one place at the start.

### DIFF
--- a/src/workers/saveToAPI.ts
+++ b/src/workers/saveToAPI.ts
@@ -87,14 +87,9 @@ const askDiff = async (
   const getCompanyBeforeAfter = () =>
     Object.keys(updated).reduce(
       ([before, after], key) => {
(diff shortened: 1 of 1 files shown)


irony (approved):
Detta behöver brytas ut framöver till separata saveEconomyToApi, saveEmissionsToApi osv. men just nu är det här tillräckligt.

PR #385, opened 2024-12-04
Post Process Paragraphs to Improve Chunking

hugo-nl:
🪂 Standalone headlines without content (body) can appear many in a row which often all refer to the body underneath the bottom headline. 
🪂 These are now merged into a paragraphs so we don't have single line paragraphs that wont provide any meaningful context and on the contrary don't have bodies without all their titles.


Greenheart (approved):
Good catch! Thanks for adding the comment to describe the intention of the code block as well - it will help us maintain this in the future.

PR #405, opened 2024-12-06
Add docling test conversion

Greenheart:
Initially using EasyOCR, but it seems like we should try using Tesseract for better performance and accuracy.


Greenheart:
Closing in favour of #406

PR #412, opened 2024-12-09
feat/recursive-search-for-company-name-if-not-found

hugo-nl:
- When it did not find the company name we got company names saying "sorry could not find company name"
- By switching to structured output we can remove that risk by recursively searching for a company name if not found
- If it is not found we throw an error


Greenheart (approved):
Great improvement! :)

PR #435, opened 2024-12-11
fix: retrying doesn't work if we use UnrecoverableError

irony:
When we try to retry because of a previous failure, the error gets rethrown as an UnrecoverableError which means it will not try again. This PR solves that problem.

Error handling improvement:

* [`src/workers/nlmParsePDF.ts`](diffhunk://#diff-c058c1bae0c644ebe0d37428b23679ce64f4063111b9a8c1e1c5b3cc441c6a24L128-R128): Changed the thrown error from `UnrecoverableError` to `Error` in the `catch` block for better general error handling.


Greenheart (approved):
Thanks for fixing this! I thought about this but didn't find time to investigate.

PR #471, opened 2024-12-12
feat/add-ids-in-the-select

hugo-nl:
- In order for the editor to work we want to be able to update/delete goals and initiatives. 
- This PR includes them from the backend giving the forntend sufficient info in order to edit specific items

### DIFF
--- a/src/routes/readCompanies.ts
+++ b/src/routes/readCompanies.ts
@@ -183,6 +183,7 @@ router.get(
           },
           goals: {
             select: {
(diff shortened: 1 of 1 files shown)


Greenheart (commented):
@hugo-nl Thanks for fixing this!

We should however keep `GET /companies` as minimal as possible. Could you move the ids to only be selected when fetching specific companies, e.g. `GET
 /companies:wikidataId`

PR #472, opened 2024-12-13
Feat: Fix deletion ability for scope1, scope2, scope1and2

hugo-nl:
- Almost ready just need some more testing

### DIFF
--- a/src/lib/prisma.ts
+++ b/src/lib/prisma.ts
@@ -23,9 +23,18 @@ const tCO2e = 'tCO2e'
 
 export async function upsertScope1(
   emissions: Emissions,
-  scope1: OptionalNullable<Omit<Scope1, 'id' | 'metadataId' | 'unit'>>,
+  scope1: OptionalNullable<Omit<Scope1, 'id' | 'metadataId' | 'unit'>> | null,
   metadata: Metadata
 ) {
+  if (scope1 === null) {
+    if (emissions.scope1Id) {
+      await prisma.scope1.delete({
+        where: { id: emissions.scope1Id },
+      })
+    }
+    return null
+  }
+
   return emissions.scope1Id
     ? prisma.scope1.update({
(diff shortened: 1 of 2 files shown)


src/lib/prisma.ts
```diff
@@ -23,9 +23,18 @@ const tCO2e = 'tCO2e'
...
 ) {
+  if (scope1 === null) {
+    if (emissions.scope1Id) {
+      await prisma.scope1.delete({
+        where: { id: emissions.scope1Id },
+      })
+    }
+    return null
+  }
+
```
Greenheart, line 37: I've thought about this. Perhaps we should use explicit `DELETE` requests instead? That way, we remove the possibility of deleting data by accident.

It's not that big of a difference between `null` and `undefined`

Greenheart: We discussed this and we will come back to this later.

src/routes/updateCompanies.ts
```diff
@@ -479,10 +479,11 @@ router.post(
...
 
       await Promise.allSettled([
-        scope1 && upsertScope1(dbEmissions, scope1, metadata),
-        scope2 && upsertScope2(dbEmissions, scope2, metadata),
+        scope1 !== undefined && upsertScope1(dbEmissions, scope1, metadata),
+        scope2 !== undefined && upsertScope2(dbEmissions, scope2, metadata),
         scope3 && upsertScope3(dbEmissions, scope3, metadata),
-        scope1And2 && upsertScope1And2(dbEmissions, scope1And2, metadata),
+        scope1And2 !== undefined &&
+          upsertScope1And2(dbEmissions, scope1And2, metadata),
```
Greenheart, line 486: Good catch! Again, let's find a more long term solution after the launch, but this should be good enough for now.

Greenheart (approved):

Issue #478, opened 2024-12-16
Simplify `StatedTotalEmissions` and remove duplicate `scope3Id` foreign key.


Greenheart:
Idea: Maybe we could remove `scope3Id` from `StatedTotalEmissions` so we only define the relationship foreign key from one side = `Scope3` and `Emissions`.

Updated `Scope3`:

```prisma
model Scope3 {
  id Int @id @default(autoincrement())

  /// Sometimes, companies only report a total value for scope 3 emissions without disclosing the scope 3 categories.
  /// Other times, they might report both, but their stated total scope 3 emissions might be different than the actual sum of their scope 3 categories.
  /// To get around this, we separate statedTotalEmissions from the actual mathematical total that we summarize during runtime.
  statedTotalEmissionsId Int? @unique
  metadataId             Int

  // The scope 3 categories, both reported and estimated.
  // TODO: Add validation so there can only be one for each category, and max 16 entries.
  categories           Scope3Category[]
  statedTotalEmissions StatedTotalEmissions? @relation(fields: [statedTotalEmissionsId], references: [id])
  emissions            Emissions?
  metadata             Metadata              @relation(fields: [metadataId], references: [id])
}
```

Updated `StatedTotalEmissions`:

```prisma
model StatedTotalEmissions {
  id         Int    @id @default(autoincrement())
  total      Float?
  metadataId Int

  unit      String
  emissions Emissions?
  metadata  Metadata   @relation(fields: [metadataId], references: [id])
  scope3    Scope3?
}
```

PR #479, opened 2024-12-16
Feat: Make totalStatedEmissions deletable

hugo-nl:
- This is important for our cleanup and editor
- However during this PR some questionable logic was found demanding a fix: #478 
- This led to a temporary fix to make sure that the statedTotalEmission id is there ( will be cleaner after 478 )


Greenheart (approved):
Good fix! Let's clean this up after the launch when we continue with the editing and validation frontend + API :)

PR #489, opened 2024-12-16
Improve the first-time local development experience for new contributors

### LINKED ISSUE #488: Make it possible to start `dev-board` without setting up `chroma`, `nlm-ingestor` and the discord bot
This would greatly simplify onboarding to work with the frontend project.

It makes no sense to require setting up the full backend project if you don't plan to develop it locally.

## Ideas to solve this:
- We could update to use dynamic `import()`, avoid importing config modules in the wrong places, etc.
- We could also maybe refactor the entry point for the board and the API to allow starting only the parts required for the API and the frontend, without importing the workers and the board.

Greenheart:
- Allow starting only the API to simplify local frontend development. This will help new contributors.
- Clarify documentation for how to start only the API, restoring a DB local backup, and testing migrations.

fix #488

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,8 +80,10 @@ This project expects some containers running in the background to work properly.
 The simplest way to start the containers the first time is to run the following docker commands.
 
 ```bash
-docker run -d -p 6379:6379 --name garbo_redis redis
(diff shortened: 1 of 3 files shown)


irony:
I would move the infra instructions for restoring DB etc to a separate .md file, maybe in the k8s folder?

PR #538, opened 2025-01-07
Schema Overhaul + API Restructuring

hugo-nl:
## Update Schema and restructure api
- In this schema we take fully advantages of the relational db where we can use cascading effects
- In order to do this we have to move the fk in lots of places so it the correct child entity cascades if the parent is deleted
- Separated concerns by using dedicated services in api
- Schema file + types file with args for DRY code
- Deletion enabled

Todo: 
1. Test everything
2. Migration script
3. Remove null option for update routes so we make sure to use deletion routes


Greenheart (commented):
Two remaining tasks before we can merge this:

1) In the migration script, we need to move the foreign key from each data point to the Metadata table. There must be a solution to both modify tables and move data at the same time, but we need to look further into how to solve this.

2) The API endpoints for deleting data have not been fully tested yet. Probably easier to finish the migration first, and then make sure they work as expected.

src/api/middlewares/middlewares.ts
```diff
@@ -1,61 +1,69 @@
...
   return (req: Request, res: Response, next: NextFunction) => {
     res.set('Cache-Control', `public, max-age=${apiConfig.cacheMaxAge}`)
     next()
   }
 }
 
+const USERS = {
+  garbo: 'hej@klimatkollen.se',
+  alex: 'alex@klimatkollen.se',
+}
```
Greenheart, line 49: This should be replaced with the `apiConfig.authorizedUsers` to save config in one place instead

Greenheart (changes requested):

Greenheart (approved):

PR #558, opened 2025-01-17
Move to cuid

hugo-nl:
### Move to cuid 
🧠 Ran in to some unexpected errors on unique constraint on goals and initiatives (during creation of new ones)
⚡ I think we might have a bad trail from a previous migration since all Id:s where unique (double-checked). 
📶 Moving to cuid removes this error (collision resistant) while preserving the existing data allowing us to move forward
🔒 Might also be better for security in the long haul

⚠️ Lets keep this PR as a draft ( even though fully functional and data preserving) since I need to discuss this with Samuel & Christian but I think this will save us some headaches for the future


Greenheart (commented):
I'd be OK to go down this route, as long as we're OK with implicitly converting old serial IDs like `42` into strings `"42"`.

However, it might not be a problem to have a database with mixed IDs where older data are numbers and newer IDs are `cuid`s.

Greenheart (approved):
Well done!

PR #560, opened 2025-01-20
Garbo with nullable fields

### LINKED ISSUE #473: Maybe use `.nullable()` instead of `.optional()` to get better responses from OpenAI
### Context:

Aha! Vissa av problemen med konstiga värden vi får tillbaka från garbo kan bero på att de inte har stöd för `.optional()` i zod-scheman, vilket vi använder på flera platser: https://github.com/openai/openai-node/issues/1180

De rekommenderar att använda `.nullable()` istället.

Men - om vi tillåter `.nullable()` och sen publicerar t.ex. https://github.com/Klimatbyran/garbo/pull/472  och https://github.com/Klimatbyran/garbo/pull/469 (redan i prod) så riskerar vi att ta bort data.

Det här talar för att vi bör skapa separata endpoints för `DELETE`-requests. Då kan vi få bättre output från garbo.

### LINKED ISSUE #522: Clarify all emissions-related prompts to never return `0` except if there is an explicit `0` in the report
This is crucial to get accurate statistics for the emissions reporting.

If no value was found, the LLM should return `null`.

### LINKED ISSUE #526: Bug in import: Sometimes empty values are resolved as `0`, which is wrong and causes problems
We should verify if values are numeric, and that the string representation of the cell value has some content.
Alternatively, check if the numeric value is 0, then the string value of the cell must also be explicitly `"0"`, otherwise we should treat the value as undefined or `null`.

Instead of resolving this in the import script, we should probably double-check in the production DB (and potentially remove) all `0` values imported via the script. We could use the timestamp and the `verifiiedBy`, and potentially `comment` to find all occurences of this problem, and fix them manually via the API / editor.

### LINKED ISSUE #531: Clarify scope 3 prompt - only interpret scope 3 if it is stated in the context

hugo-nl:
Nullable Garbo > Garbo 💪 

#### 🌍 Initial Tests Show Great Promise
Garbo’s ability to hallucinate and fabricate values has been a significant pain point. With this update, Garbo now explicitly sets missing values to null instead of making them up, aligning better with the principle of accurate and transparent data handling.

🤖 Initially, I updated the API to accept null values without treating them as changes. However, upon reviewing the impact on scalar fields, I realized having nullable fields could potentially confuse API users. To address this, I implemented a helper function: removeNullValuesFromGarbo, which ensures that only actual payload data is sent to the API. This way, the API remains clean and intuitive without requiring alterations.

#### ⚡ Next Steps
The removeNullValuesFromGarbo function has been tested across a variety of cases and has shown great results. I’m running a few final reports to ensure robustness, but this improvement already feels like a significant step forward in resolving one of Garbo’s most persistent issues.

Closes #473
Closes #531
Closes #526
Closes #522


src/workers/saveToAPI.ts
```diff
@@ -36,9 +36,36 @@ export const saveToAPI = new DiscordWorker<SaveToApiJob>(
...
+          )
+
+          return Object.keys(sanitizedObject).length > 0
+            ? sanitizedObject
+            : null
+        } else {
+          return data
+        }
+      }
+
```
Greenheart, line 63: This function could remove `scope3.categories` if it is an empty array. This would allow us to remove the entire `scope3`object if no proper data was found.

Greenheart (commented):
This seems really promising!

src/workers/saveToAPI.ts
hugo-nl: Stated-total-emissions might still be valid though?

hugo-nl: Having garbo use .nullable for teh entire array should solve it!

Greenheart: Agreed, `.nullable()` for the array should be good. And statedTotalEmissions should be OK too.

src/prompts/followUp/economy.ts
```diff
@@ -8,17 +8,20 @@ export const schema = z.object({
         .object({
           turnover: z
             .object({
-              value: z.number().optional(),
-              currency: z.string().optional(),
+              value: z.number().nullable().optional(),
+              currency: z.string().nullable().optional(),
```
Greenheart, line 12: Using both `nullable()` and `optional()` might look weird, but this is actually because OpenAI doesn't support `optional()` for their structured output schemas. Thus we need to add `nullable()`. However, we agreed to keep both for now if/when `optional()` becomes an alternative.

Greenheart (approved):
Good job! This will solve many long-standing issues :)

PR #563, opened 2025-01-21
Fix scalar empty body error

hugo-nl:
⚙️ I noticed that I got 500 errors when testing deletion from scalar
⚡ The reason is that our deletion routes are set to be application/json in scalar which means they demand a json body
🌍 This fix sets a default body "{}" which makes the api documentation work as intended for deletions


Greenheart (commented):
I suggest we avoid merging this PR until we know if/how Scalar addresses the actual underlying issue in their API client. See https://github.com/scalar/scalar/issues/4542 for more details :)

PR #567, opened 2025-01-21
Return Ids for editorial purposes

hugo-nl:
🌍 The editor needs to be able to perform CRUD ops from the client --> Needs ids to know what items to change
🧹 Cleaned up our types/args/schema code a bit


src/api/routes/company.read.ts
```diff
@@ -314,143 +175,7 @@ export async function companyReadRoutes(app: FastifyInstance) {
...
-                year: true,
-                scope: true,
-                metadata,
-              },
-              orderBy: {
-                year: 'desc',
-              },
-            },
-          },
+          ...detailedCompanyArgs,
```
Greenheart, line 178: This works as expected, though it would be better to always spread before setting custom properties. In this case the `where` clause is set before we spread the rest of the args, which (in this case) don't include anything that would overwrite the `where` clause defined in the API endpoint.

It would be good to keep all spread operators at the start of each object to avoid ambiguous bugs.

src/api/schemas.ts
```diff
@@ -182,10 +182,13 @@ export const postReportingPeriodsSchema = z
   })
   .merge(createMetadataSchema)
 
+// Response Schemas Below
```
Greenheart, line 185: This might be overlooked or become outdated. Also, the module is 500+ lines. What if we split the API schemas into the following modules?

This would be a quick fix to make the schemas easier to understand and maintain.

```txt
src/api/schemas/
    common.ts
    request.ts
    response.ts
```

Greenheart (changes requested):
Some good improvements in here!

I think a quick iteration based on the comments could make this part of the code even better.

src/api/schemas.ts
hugo-nl: Agree! Fixing 👍

Greenheart (approved):
Looking good with the latest changes! Ready to merge :)

PR #568, opened 2025-01-22
Update production with latest changes: Nullable fields + returning Ids  and schema/types cleanup

hugo-nl:


Greenheart (approved):
Ready to deploy :)

PR #571, opened 2025-01-22
Update and simplify error handling

### LINKED ISSUE #562: Expose validation errors to the client
In the error handler (https://github.com/Klimatbyran/garbo/blob/df9f53636ce631ad7725a3a4797097706432a3d4/src/api/plugins/errorhandler.ts), we should check `if (error instanceof ZodError)` and in that case return the validation error.

The error is more likely coming from the schema validation library, so these could be the actual error types we want to check against https://github.com/turkerdev/fastify-type-provider-zod/pull/110/files

Fixing this would make it easier to work with the API and debug invalid requests.

However, we should only expose the validation requests, and not the serialisation errors, since those are internal implementation details.

## To implement this:
- [ ] Check for Fastify request validation errors in `errorhandler.ts`
- [ ] return `reply.status(400).send(validationError)` from `errorhandler.ts`
- [ ] Use the standard Fastify validation error response and just make a schema for it which can be added to all API endpoints where it could occur. Or - maybe we want to add this as default to all endpoints? There must be a way to document that in OpenAPI / Fastify / `@fastify/swagger`? If we can define it centrally, it would save a lot of duplicated imports. Though it is nice to explicitly define the expected errors for every API endpoint.

hugo-nl:
🚋 Centralized error middleware --> Removing `GarboAPIError ` 🗑️ 
🌍 Adding fastify validation errors

Closes #562


Greenheart (changes requested):
We should define the schemas for HTTP 400 errors to make it explicit in the OpenAPI docs. This would make it easier to handle validation errors, and know what to expect when developing the frontend.

Greenheart (approved):
This turned out pretty nice! Ready to merge.

PR #572, opened 2025-01-23
Feat/municipality data in API

### LINKED ISSUE #557: Add GET endpoints to load municipality data via the API
This allows us to load all data from the same place.

If we add the municipality data to the API, we could get better OpenAPI docs which would make it easier to get AI:s to generate frontend code using the expected data.

This would also make it easier for others outside of Klimatkollen to use the open datasets that we've compiled and processed.

## Initial tasks
- [x] `GET /municipalities` - list all `climate-data.json`
- [x] `GET /municipalities/:municipalitySlug` (or ID) - details for specific municipality

## Implementation ideas
- Add https://github.com/Klimatbyran/klimatkollen/blob/staging/data/output/climate-data.json to garbo
- Create a `municipalityService` which supports `getMunicipalities()` and `getMunicipality(name: string)`
- In the `municipalityService`, load `climate-data.json` and keep a `Map` in memory to allow returning company data quickly.
- Implement the `cache()` plugin to make the client save responses.
- Generate zod schemas based on the JSON data, for example with https://app.quicktype.io/. Use these both to parse the incoming data from the Python data pipeline when garbo starts, to ensure we get the expected data format. Also define the same schemas for the API endpoints response schemas. This will also add this to the API documentation.

Greenheart:
## Add municipality data to the API

- `GET /municipalities` returns a list of all entries.
- `GET /municipalities/:name` returns a specific entry.

The `municipalityService` is lazy-loading data the first time it is requested, which reduces the startup time and memory usage until the data is actually needed. In addition we also send regular cache headers to the client.

The `climate-data.json` rarely changes and the file is part of this git repo. In the future we also want to move the Python data pipeline here too.

This PR also improves the type for `getErrorSchemas()`

---

## What remains to be done before merging
- I think we could improve the OpenAPI docs with better example responses for the `200` responses. I've got some ideas.

Resolves #557


src/api/routes/municipality.read.ts
```diff
@@ -0,0 +1,68 @@
...
+      reply
+    ) => {
+      const { name } = request.params
+      const municipality = municipalityService.getMunicipality(name)
+
+      if (!municipality) {
+        return reply.status(404).send({
+          code: 'NOT_FOUND',
+          message: 'The requested resource could not be found.',
+        })
```
Greenheart, line 62: Should we keep common errors in one place? Perhaps a module `api/errors.ts` from where we export common errors, like `404`. Or this could be a function

```ts

const errors = {
  404: {
    code: 'NOT_FOUND',
    message: 'The requested resource could not be found.
  }
}

export function sendError(statusCode: 404 | (...), reply: FastifyReply) {
  return reply.status(statusCode).send(errors[statusCode])
}
```

Since this is no Prisma error, we can't handle it as a regular 404 error. This is the kind of situation where `GarboAPIError` is useful, to allow us to send a common error message.

Compared to using a function, the `GarboAPIError` would achieve the same affect, since that would also abort the request handler early, and return a default error response in the `errorhandler.ts`.

src/api/schemas/response.ts
```diff
@@ -337,3 +337,34 @@ export const CompanyDetails = CompanyBase.extend({
...
+  electricCarChangeYearly: z.record(z.string(), z.number()),
+  climatePlanLink: z.string(),
+  climatePlanYear: z.union([ClimatePlanYearEnumSchema, z.number()]),
+  climatePlanComment: z.string(),
+  bicycleMetrePerCapita: z.number(),
+  totalConsumptionEmission: z.number(),
+  electricVehiclePerChargePoints: z.number(),
+  procurementScore: z.string(),
+  procurementLink: z.string(),
+})
```
Greenheart, line 368: This schema could use a better example response in the OpenAPI docs. That would make it easier to know what to expect on the client side.

I have some ideas on how to solve this.

irony:
Good stuff! We should rename the fields either in the data or through a schema refine? So the official api is not mixing Swedish and English. (Län => region, namn => name)

src/api/routes/municipality.read.ts
irony: Put that as a separate issue

src/api/schemas/response.ts
irony: Län and kommun needs to be translated to english

Klimatkollen: Kommun = municipality.
Län is a bit harder. The official Swedish translation into English is "county", and I think this is what we should use. However, it does not work so well in a US context, because a county in USA is closer to a Swedish "kommun". Other related terms are "state", "province" or "region", but none of them are really correct for the Swedish "län".

Greenheart: Yeah, the best place for this would be in the Python data pipeline, since we have control of the entire flow.

I think we should use `region` as a translation for `län`, because it's both a generic term that works internationally, while being the technically most correct term in swedish as well:
https://sv.wikipedia.org/wiki/Region_(landsting)

hugo-nl (approved):
Super nice, great work! 🙌 👍

PR #582, opened 2025-01-24
Improve municipality response format to use explicit
year/value pairs

### LINKED ISSUE #578: Improve response format of `GET /municipalities` and `GET /municipalities/:name` to return explicit names to reduce risk for LLM code gen errors
As an alternative to #574, we could improve the response API response format to be more explicit about what the data means.

For example, where we currently return this:

```json
{
    "emissions": {
      "1990": 396555.825518647,
      "2000": 291563.906229046,
      "2005": 301791.034669901,
      "2010": 306953.831737827,
      "2015": 237247.290285871,
      "2016": 236423.864824126,
      "2017": 233451.264481917,
      "2018": 229427.286373341,
      "2019": 227117.00904579,
      "2020": 211623.571089559,
      "2021": 211715.680273485,
      "2022": 195210.46458019
    }
}
```

We could instead return this:

```json
{
    "emissions": [
      { "year": 1990, "value": 396555.825518647 },
      { "year": 2000, "value": 291563.906229046 },
      { "year": 2005, "value": 301791.034669901 },
      { "year": 2010, "value": 306953.831737827 },
      { "year": 2015, "value": 237247.290285871 },
      { "year": 2016, "value": 236423.864824126 },
      { "year": 2017, "value": 233451.264481917 },
      { "year": 2018, "value": 229427.286373341 },
      { "year": 2019, "value": 227117.00904579 },
      { "year": 2020, "value": 211623.571089559 },
      { "year": 2021, "value": 211715.680273485 },
      { "year": 2022, "value": 195210.46458019 }
    ]
}
```

Since we give clear names for each key (`year`), I think this would improve accuracy for the LLM code generation, since it might know how to use the `year` and refer the data to the prompts.

This would also improve the auto-generated examples in the API documentation.

## Potential drawbacks:

1) It's slightly more verbose and requires more data to be transferred to the client. But this won't really matter with proper caching since this data will mostly be fetched once per client.

2) This approach also loses the ability to lookup data for a given year directly by key. This could easily be solved by adding a lookup problem during runtime with a lookup object if we need it.

Since neither is a valid reason - we could go ahead with this :)

irony:
## Problem:
The current municipality endpoints return yearly data as key-value objects,
which can be ambiguous for LLM code generation and API documentation. For
example:

```json
{
  "emissions": {
    "1990": 396555.82,
    "2000": 291563.90
  }
}
```

## Solution:
Transform the response to use explicit year/value pairs in arrays, making
the data structure more self-documenting:

```json
{
  "emissions": [
    { "year": "1990", "value": 396555.82 },
    { "year": "2000", "value": 291563.90 }
  ]
}
```

This change:

  - Makes the data structure more explicit and easier to understand
  - Improves OpenAPI documentation with better examples
  - Helps LLMs better understand and generate code for the API
  - Adds validation to ensure years are in YYYY format

The changes affect both `municipalities` and `/municipalities/:name\`
endpoints.

Fixes #578


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "garbo",
-  "version": "3.6.0",
+  "version": "3.6.7",
```
irony, line 3:
```suggestion
  "version": "3.6.0",
```

Greenheart (approved):
Thanks for starting this!

Unfortunately, these code changes originally didn't work as expected, so I fixed the errors by separating the input file format from API response format.

I also made the transformation faster by only doing it the first time the municipality data is requested - compared to repeating the same transformations for every request.

This now works as expected both for `GET /municipalities` and `GET /municipalities/:name`.

PR #587, opened 2025-01-27
Add redis caching

hugo-nl:
#### Add Redis Caching

🌍  This PR introduces Redis-based caching for better performance and reduced database load.

Key Updates:
	🌍 Companies API: 
                 - Implements Redis cache with dynamic keys tied to latestMetadataUpdatedAt.
                 - Reduces redundant database queries by serving cached data when available.
	🌍 Municipalities API:
                 - Adds Redis cache with a static key since data is immutable.
                 - Enhances performance with faster responses for repeated requests.

Benefits:
        🌍 Optimized server performance.
	🌍 Simplified cache management with automatic invalidation for updates.
	🌍 Scalable design leveraging Redis.
	
🏠 Updated the readme with a quick setup for redis insight. Where you can inspect you redis data 

Testing:
	✅ Verified caching works for /companies and /municipalities APIs.
	✅ Ensured no impact on existing ETag implementation.

### DIFF
--- a/src/api/schemas/response.ts
+++ b/src/api/schemas/response.ts
@@ -14,7 +14,9 @@ export const MetadataSchema = z.object({
     .nullable()
     .openapi({ description: 'Comment about the data' }),
   source: z.string().nullable().openapi({ description: 'Source of the data' }),
-  updatedAt: z.date().openapi({ description: 'Last update timestamp' }),
+  updatedAt: z
+    .union([z.date(), z.string()])
+    .openapi({ description: 'Last update timestamp' }),
   user: z.object({
     name: z
       .string()
@@ -224,9 +226,11 @@ export const InitiativeSchema = z.object({
 export const ReportingPeriodSchema = z.object({
   id: z.string(),
   startDate: z
-    .date()
+    .union([z.date(), z.string()])
     .openapi({ description: 'Start date of reporting period' }),
-  endDate: z.date().openapi({ description: 'End date of reporting period' }),
+  endDate: z
+    .union([z.date(), z.string()])
+    .openapi({ description: 'End date of reporting period' }),
   reportURL: z
     .string()
     .nullable()
--- a/src/api/schemas/response.ts
+++ b/src/api/schemas/response.ts
@@ -4,6 +4,13 @@ import { wikidataIdSchema } from './common'
 
 extendZodWithOpenApi(z)
 
+const dateStringSchema = z.union([
+  z.date(),
+  z.string().refine((val) => !isNaN(Date.parse(val)), {
+    message: 'Invalid date string',
+  }),
+])
+
 export const okResponseSchema = z.object({ ok: z.boolean() })
 export const emptyBodySchema = z.undefined()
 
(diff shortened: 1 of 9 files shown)


src/api/schemas/response.ts
```diff
@@ -224,9 +226,11 @@ export const InitiativeSchema = z.object({
 export const ReportingPeriodSchema = z.object({
   id: z.string(),
   startDate: z
-    .date()
+    .union([z.date(), z.string()])
```
Greenheart, line 229: I assume the zod union is needed here to work with the cached data from Redis, which might be JSON data?

If we use `z.string().date()`, we could make sure to only return valid date strings.

src/api/schemas/response.ts
```diff
@@ -224,9 +226,11 @@ export const InitiativeSchema = z.object({
...
     .openapi({ description: 'Start date of reporting period' }),
-  endDate: z.date().openapi({ description: 'End date of reporting period' }),
+  endDate: z
+    .union([z.date(), z.string()])
```
Greenheart, line 232: Same thing here with the `z.string().date()`

src/createCache.ts
```diff
@@ -1,31 +1,29 @@
...
-    delete(key: string) {
-      cache.delete(key)
+    async delete(key: string) {
+      await redis.del(key)
     },
-    clear() {
-      cache.clear()
+    async clear() {
+      const keys = await redis.keys('*')
+      if (keys.length > 0) await redis.del(keys)
```
Greenheart, line 26: This could be really dangerous. Since we use the same Redis instance for BullMQ, we should only let the API the delete the API cache keys.

What if we used a prefix for our the API cache keys to clearly separate the API cache from the BullMQ state?

src/createCache.ts
```diff
@@ -1,31 +1,29 @@
...
+const redis = createClient({ ...redisConfig })
+redis.connect()
 
+export function createServerCache({ maxAge }: { maxAge: number }) {
   return {
-    set(key: string, value: any) {
-      cache.set(key, { data: value, cachedAt: Date.now() })
+    async set(key: string, value: any) {
+      await redis.set(key, JSON.stringify(value), {
+        PX: maxAge,
```
Greenheart, line 11: Excellent to implement automatic clearing!

src/api/schemas/response.ts
```diff
@@ -14,7 +14,9 @@ export const MetadataSchema = z.object({
     .nullable()
     .openapi({ description: 'Comment about the data' }),
   source: z.string().nullable().openapi({ description: 'Source of the data' }),
-  updatedAt: z.date().openapi({ description: 'Last update timestamp' }),
+  updatedAt: z
+    .union([z.date(), z.string()])
```
Greenheart, line 18: same thing with `z.string().date()`

Greenheart (changes requested):
Overall a good improvement!

Some important things to consider and resolve before merging.

src/api/schemas/response.ts
hugo-nl: Unfortunately it gives some when the data is actually a date. I think this below is a safe way to do it. 

const dateStringSchema = z.union([
  z.date(),
  z.string().refine((val) => !isNaN(Date.parse(val)), {
    message: 'Invalid date string',
  }),
])

src/api/schemas/response.ts
hugo-nl: 👆

src/createCache.ts
hugo-nl: Nice catch! Fixed with a cache prefix 👍

src/createCache.ts
hugo-nl: 🙌 🙌

Greenheart (approved):
Thanks for these improvements! Good thing we fixed the cache key before deploying :smile:

PR #588, opened 2025-01-27
Fix schema fk for economy

hugo-nl:
🌍 Removed redundant foreign key: Eliminated the excessive fkTurnoverId from the Economy schema since economyId already exists on the Turnover entity.

🌍 Fixed reporting period upsert bug: Simplified and corrected the logic for upsertReportingPeriod.

### DIFF
--- a/prisma/schema.prisma
+++ b/prisma/schema.prisma
@@ -256,7 +256,6 @@ model Economy {
   turnover        Turnover?
   employees       Employees?
   reportingPeriod ReportingPeriod? @relation(fields: [reportingPeriodId], references: [id], onDelete: Cascade)
-  turnoverId      String?
 }
 
 model Turnover {
--- a/src/api/services/reportingPeriodService.ts
+++ b/src/api/services/reportingPeriodService.ts
@@ -18,18 +18,23 @@ class ReportingPeriodService {
       year: string
     }
   ) {
-    const reportingPeriod = await prisma.reportingPeriod.findFirst({
-      where: {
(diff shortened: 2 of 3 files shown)


src/api/services/reportingPeriodService.ts
```diff
@@ -18,18 +18,23 @@ class ReportingPeriodService {
...
       where: {
-        id: reportingPeriod?.id ?? '',
+        companyId_year: {
+          companyId: company.wikidataId,
+          year,
+        },
+      },
+      update: {
+        startDate,
+        endDate,
```
Greenheart, line 30: This works in most cases. However, I think the reason we kept the `update` block empty was that we most of the time wanted to `create` the reporting period and/or return it, but not necessarily update it at the same time.

We could try this approach and hopefully it works going forward. But if you get future problems with garbo changing the reportingPeriod `startDate` and/or `endDate` (potentially causing conflicts if there are multiple reports for the same year), then consider separating the endpoints for creating and updating reportingPeriods.

prisma/schema.prisma
```diff
@@ -256,7 +256,6 @@ model Economy {
   turnover        Turnover?
   employees       Employees?
   reportingPeriod ReportingPeriod? @relation(fields: [reportingPeriodId], references: [id], onDelete: Cascade)
-  turnoverId      String?
```
Greenheart, line 259: Ah, good catch!

Greenheart (approved):
This works well locally :+1:

The comment is mainly for information and historical context.

PR #590, opened 2025-01-27
Fix employee units

### LINKED ISSUE #383: Update `economy` prompt to clarify unit for `employees`
Several reports got `FTE` as the unit, despite actually reporting the number of employees as `Average`, `EOY` (End of Year), or `Head count`.

This is a nice to have, but not critical right now.

### LINKED ISSUE #463: Weird units, in English, for employees for companies ('head count' etc)
Unit for number of company employees are strange and in English. Examples as 'average number of employees during the year' (Q131413576) or 'head count' (Q1571428). Should be in Swedish and suggestion is that FTE should be baseline unit (but presented as 'personer'/'stycken'/no unit on website).

hugo-nl:
🌍 Updated employee unit handling in Garbo:
	•	Added other acceptable units (e.g., AVG, EOY) to avoid faulty usage of FTE.
	•	Clarified instructions to set the unit to null if no unit is specified.
	

Closes #383  
Closes #463

### DIFF
--- a/src/prompts/followUp/economy.ts
+++ b/src/prompts/followUp/economy.ts
@@ -16,7 +16,7 @@ export const schema = z.object({
           employees: z
             .object({
               value: z.number().nullable().optional(),
-              unit: z.string().nullable().optional(),
+              unit: z.enum(['FTE', 'EOY', 'AVG']).nullable().optional(),
             })
             .nullable()
             .optional(),
@@ -45,7 +45,7 @@ export const prompt = `
 - Acceptable units include:
   - FTE (Full-Time Equivalent)
   - AVG (genomsnittligt antal anställda)
-  - EOY (Antal anstälda vid årets slut)
+  - EOY (Antal anställda vid årets slut)
 - If no unit is specified, extract the value as is and set the unit field to null.
 *** Dates: ***
 - if no year is specified, assume the current year ${new Date().getFullYear()}
--- a/src/prompts/followUp/economy.ts
+++ b/src/prompts/followUp/economy.ts
@@ -45,8 +45,7 @@ export const prompt = `
 - Acceptable units include:
   - FTE (Full-Time Equivalent)
   - AVG (genomsnittligt antal anställda)
-  - EOY (End of Year)
-  - Headcount
+  - EOY (Antal anstälda vid årets slut)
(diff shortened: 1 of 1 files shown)


src/prompts/followUp/economy.ts
```diff
@@ -41,7 +41,12 @@ export const prompt = `
   - 4.2 KEUR → 4200 EUR
 - Specify the **currency** as a separate field (e.g., SEK, USD, EUR).
 *** Employees: ***
-- Extract the number of employees for all available years. The unit can be for example "FTE" (full-time equivalent) or average number of employees during the year.
+- Extract the number of employees for all available years. 
+- Acceptable units include:
+  - FTE (Full-Time Equivalent)
+  - AVG (genomsnittligt antal anställda)
+  - EOY (Antal anstälda vid årets slut)
```
Greenheart, line 48:
```suggestion
  - EOY (Antal anställda vid årets slut)
```

Greenheart (changes requested):
Would it be possible to specify this value as an enum in the schema as well?

OpenAI structured output supports enum values:
https://platform.openai.com/docs/guides/structured-outputs/supported-types#supported-types

This means we might be able to use something like `z.enum(['FTE', 'AVG', 'EOY'])` to keep the units consistent.

JSON Schema docs:
https://json-schema.org/understanding-json-schema/reference/enum

src/prompts/followUp/economy.ts
hugo-nl: Nice catch, fixed!

hugo-nl:
> Would it be possible to specify this value as an enum in the schema as well?
> 
> OpenAI structured output supports enum values: https://platform.openai.com/docs/guides/structured-outputs/supported-types#supported-types
> 
> This means we might be able to use something like `z.enum(['FTE', 'AVG', 'EOY'])` to keep the units consistent.
> 
> JSON Schema docs: https://json-schema.org/understanding-json-schema/reference/enum

Yes I thought about this. I ended up not doing it in case I forgot a valid unit. But if we think these above should do it then I am all for it. What do you think?

Greenheart (approved):
Good improvements! :)

If we need more units, that will be easy to fix then.

PR #592, opened 2025-01-28
Update Scalar to respect empty body for `DELETE` requests

### LINKED ISSUE #591: Update Scalar to respect empty body for `DELETE` requests
Since https://github.com/scalar/scalar/pull/4550 is now fixed, we could update Scalar to verify the issue has been resolved. This will make it easier to test and work with `DELETE` requests.

Greenheart:
This makes it much easier to delete companies or specific data points. Very useful during development since this reduces the need for resetting the entire database.

Fixes #591


hugo-nl (approved):
Very cool that they resolved it for us! Nice win 🙌

PR #594, opened 2025-01-28
Add enum option to support tCO2 when tCO2e suffix is not available

### LINKED ISSUE #533: Add support in garbo prompt and schema for handling different units like `tCO2`, `tCO2e`
For example, the Swedavia report for 2023 used `tCO2` as the unit instead of `tCO2e` like almost all other reports. The reason why this is important is because other companies may include other kinds of GHG emissions, while Swedavia 2023 only disclosed their CO2 emissions. This means their total `CO2e` emissions might be higher than if you only include the `CO2` emissions. Thus, we can not simply convert `1 tCO2 === 1 tCO2e`. Or can we? What do we do if we have other units that can't easily be converted into `tCO2e`?

We have support for specifying the `unit` in the database, but not in the garbo prompts and schemas.

The frontend will be updated separately to use the actual units, rather than a hard coded `tCO2e` like we could do in the beginning.

hugo-nl:
🌍 Updated schema to support both tCO2e (preferred) and tCO2 as valid units for emissions reporting.
🌍 Ensured tCO2 is only used when explicitly reported without the “e” suffix.
🌍 Included unit conversion for tCO2 and tCO2e to ensure all values are consistently reported in metric tons.

Closes #533 
Closes: #569

### DIFF
--- a/src/api/schemas/request.ts
+++ b/src/api/schemas/request.ts
@@ -1,5 +1,5 @@
 import { z } from 'zod'
-import { wikidataIdSchema } from './common'
+import { emissionUnitSchema, wikidataIdSchema } from './common'
 
 const createMetadataSchema = z.object({
   metadata: z
@@ -74,7 +74,7 @@ export const postIndustrySchema = z
   .merge(createMetadataSchema)
 
 export const statedTotalEmissionsSchema = z
-  .object({ total: z.number(), unit: z.enum(['tCO2e', 'tCO2']) })
+  .object({ total: z.number(), unit: emissionUnitSchema })
   .optional()
 
 export const emissionsSchema = z
@@ -82,7 +82,7 @@ export const emissionsSchema = z
     scope1: z
       .object({
         total: z.number(),
-        unit: z.enum(['tCO2e', 'tCO2']),
+        unit: emissionUnitSchema,
       })
       .optional(),
     scope2: z
@@ -96,7 +96,7 @@ export const emissionsSchema = z
         unknown: z
           .number({ description: 'Unspecified Scope 2 emissions' })
           .optional(),
-        unit: z.enum(['tCO2e', 'tCO2']),
+        unit: emissionUnitSchema,
       })
       .optional(),
     scope3: z
@@ -106,7 +106,7 @@ export const emissionsSchema = z
             z.object({
               category: z.number().int().min(1).max(16),
               total: z.number(),
-              unit: z.enum(['tCO2e', 'tCO2']),
+              unit: emissionUnitSchema,
             })
           )
           .optional(),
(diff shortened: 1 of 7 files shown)


src/api/schemas/request.ts
```diff
@@ -74,14 +74,15 @@ export const postIndustrySchema = z
   .merge(createMetadataSchema)
 
 export const statedTotalEmissionsSchema = z
-  .object({ total: z.number() })
+  .object({ total: z.number(), unit: z.enum(['tCO2e', 'tCO2']) })
```
Greenheart, line 77: Since the units are common and shared for all emission values, could we move these to `src/api/schemas/common.ts`?

src/api/schemas/request.ts
```diff
@@ -95,15 +96,8 @@ export const emissionsSchema = z
         unknown: z
           .number({ description: 'Unspecified Scope 2 emissions' })
           .optional(),
+        unit: z.enum(['tCO2e', 'tCO2']),
```
Greenheart, line 99: Should the `unit` be required for all emission values? Or should we make `unit` optional everywhere and assume `tCO2e` by default? We should also note that the Zod enum is case-sensitive.

It makes sense to force garbo to return the unit. But for the editor (which is using the same API endpoint to save emissions), we might not want to have to pass along the unit with every request.

By forcing explicit units for all data points, we might get better data, but it could also be limiting for the clients if we only allow specific units. Since we currently only expect updates from our own client, where we use `openapi-fetch` to stay in sync with the expected enum values, this  would not be a problem for us.

hugo-nl: Good thinking! Fix--> Default zod value which makes it optional

src/api/schemas/request.ts
```diff
@@ -95,15 +96,8 @@ export const emissionsSchema = z
...
+        unit: emissionUnitSchema,
       })
-      .refine(
-        ({ mb, lb, unknown }) =>
-          mb !== undefined || lb !== undefined || unknown !== undefined,
-        {
-          message:
-            'At least one property of `mb`, `lb` and `unknown` must be defined if scope2 is provided',
-        }
-      )
```
Greenheart, line 106: This validation would still be useful for the API

Greenheart (changes requested):

src/api/services/emissionsService.ts
```diff
@@ -12,7 +12,7 @@ import { DefaultEmissions } from '../types'
 import { prisma } from '../../lib/prisma'
 import { emissionsArgs } from '../args'
 
-const TONNES_CO2_UNIT = 'tCO2e'
+export const TONNES_CO2E_UNIT = 'tCO2e'
```
Greenheart, line 15: Maybe this should be the default unit used for the `emissionUnitSchema`? If we move this constant to the same module as the `emissionUnitSchema`, we could also keep related values together.

Or should this be a part of the config? I'm happy with keeping it in the common schemas to keep it simple.

Greenheart (approved):
Thanks! Good improvements in the latest commits :)

PR #595, opened 2025-01-28
Exclude Scope 3 Category 16 Emissions Unless Verified

### LINKED ISSUE #527: If `scope3.calculatedTotalEmission` is based on `scope3.categories`, we should exclude category `16: other` unless it has been verified.
This way, we could avoid invisible numbers affecting the `calculatedTotalEmissions`.

To fix this, we should update `company.read.ts`. Or search for `calculatedTotalEmissions` to find it.

Remember to update for both endpoints: `/companies` and `/companies/:wikidataId`

hugo-nl:
🌍 Updated calculatedTotalEmissions logic to exclude Scope 3 Category 16 (other) unless verified, ensuring invisible numbers do not affect calculations.

Closes #527


Greenheart (approved):
Good fix!

PR #611, opened 2025-02-11
Improved readme by mentioning potential missing dependencies and spli…

moritzzingg-netlight:
Improved two things in the readme:

- Split up the commands to initially start the docker containers to make it a bit faster to copy them using the github command copy widgets
- Added an additional note regarding dependencies on unix-based system. PDF2Image requires some additional packages that might need to be installed.


hugo-nl (approved):
Super nice! Readme win 🙌

PR #612, opened 2025-02-11
Feat: Added docker compose file

okis-netlight:
Added a docker compose file to simplify container startup. Updated readme accordingly.


hugo-nl (approved):
Works great! Way easier to start up garbo this way 🥇

PR #614, opened 2025-02-12
feat: tags are now included when fetching a list of all companies

### LINKED ISSUE #613: Return the tags to the frontend
- In order to filter on the frontend we could use the tags
- In the company.read file we return the companies
- Update response schemas accordingly

Image

okis-netlight:
Fixes #613 

- Updated response body schema for getting all companies
- Updated db get arguments to include tags


hugo-nl (approved):
Well done! 🤝 🙌

Issue #539, opened 2025-01-07
Implement GitHub authentication


Greenheart:
This should connect a GitHub account (`userId`) to a user in our database, to allow us to track metadata for the change history.

With GitHub authentication in place, we should have some way to distinguish trusted users - essentially some kind of authorization, e.g. an role-based access control system where each account can be given a number of roles, and only for example users with the `editor` role may edit our climate data.

We also want to manage which users we trust as editors or viewers. This could be done via GitHub organization teams, to manage members. Then we could have a `Editor` team and an `Viewer` team (who can see non-public data).

moritzzingg-netlight:
With the merge of the corresponding PRs this issue should be completed

PR #644, opened 2025-02-25
[Feat] Implement GitHub authentification

moritzzingg-netlight:
Integrated proper authentication into the backend using Github as an IDP 
- Implemented necessary endpoint to exchange Github access token for a garbo API access token
- API Tokens are implemented using JWT
- Expanded the user model
- Modified the auth plugin to validate API access tokens

Closes backend part of #539


src/api/services/authService.ts
```diff
@@ -0,0 +1,88 @@
...
+                email: userinfo.email
+            }
+        })   
+        
+        return this.createToken(user);
+    }
+
+    verifyUser(token: string) {
+        const payload = jwt.verify(token, apiConfig.jwtSecret) as User & {exp: number};
+        if(Date.now() > payload.exp - 15 * 60) {
```
hugo-nl, line 77: Could we put this into a constant? Better for readability adn does date.now() come in millisekonds?

I would do something like this:

```
  private static readonly TOKEN_EXPIRY_BUFFER_MINUTES = 15;
  private static readonly SECONDS_IN_A_MINUTE = 60;
  
const currentTimeInSeconds = Date.now() / 1000;
const renewalThreshold = AuthService.TOKEN_EXPIRY_BUFFER_MINUTES * AuthService.SECONDS_IN_A_MINUTE;

    const shouldRenew = currentTimeInSeconds > payload.exp - renewalThreshold;

    return {
      user: payload,
      newToken: shouldRenew ? this.createToken(payload) : undefined
    };
```

src/api/schemas/request.ts
```diff
@@ -189,3 +189,7 @@ export const MunicipalityNameSchema = z.string()
 export const MunicipalityNameParamSchema = z.object({
   name: MunicipalityNameSchema,
 })
+
+export const authentificationBodyScheme = z.object({
```
hugo-nl, line 193: Typo :)

hugo-nl (changes requested):
## Nicely done Moritz! 👏 🎉 
🌎 This PR was super impressive, I had a few nit picks but then we are ready to ship

src/api/schemas/request.ts
moritzzingg-netlight: Fixed

src/api/services/authService.ts
moritzzingg-netlight: Fixed

hugo-nl (approved):
## Lets go! 🚢 🚢 🚢

PR #647, opened 2025-02-28
fix: Handle API tokens and JWT tokens separately in auth plugin

irony:
After examining the code, I found the issue in the auth plugin. The problem is in how the token is being verified. The
code is trying to verify API tokens using JWT verification, but these appear to be two different authentication
mechanisms.

* Improved token verification logic by separating the handling of simple API tokens from JWT tokens.
* Added a check to see if the token is a simple API token and, if so, set `request.user` to null and return early.
* Simplified the error logging when no token is provided by removing unnecessary details.


moritzzingg-netlight:
Should be solved by #656

Issue #648, opened 2025-02-28
Remove default value for JWT secret


okis-netlight:
https://github.com/Klimatbyran/garbo/blob/afd5209f5c0a1dae724b1a6607984c7396788ae4/src/config/api.ts#L31

When not setting the JWT_SECRET environment variable, this results in our config using the default value here. Since this value is exposed in GitHub it is not safe to use. To solve, remove the default value, which will not allow for this type of misconfiguration.

elvbom:
Seems to be done so closing

PR #649, opened 2025-02-28
Remove default values for sensitive environment variables

okis-netlight:
Solves #648 

Removed default values as to not allow any configurations with missing environment variables to use secrets exposed in the source code.


okis-netlight:
Will be fixed in #690

PR #656, opened 2025-03-06
Add /auth/token endpoint for service to exchange secret for api token

willeasp:
# Fix client missing client authentication
- Add /auth/token endpoint for services, users still use /auth/github to retrieve tokens
- Introduce a secret (stored as ENV variable `API_SECRET` currently)
- introduce `bot: boolean` field in User entity

### DIFF
--- a/src/api/plugins/auth.ts
+++ b/src/api/plugins/auth.ts
@@ -1,8 +1,6 @@
 import { FastifyInstance, RouteGenericInterface } from 'fastify'
 import fp from 'fastify-plugin'
 import { User } from '@prisma/client'
-
-import apiConfig from '../../config/api'
 import { authService } from '../services/authService'
 
 declare module 'fastify' {
@@ -10,8 +8,7 @@ declare module 'fastify' {
     user: User | null
   }
 
-  export interface AuthenticatedFastifyRequest<T extends RouteGenericInterface>
-    extends FastifyRequest<T> {
+  export interface AuthenticatedFastifyRequest<T extends RouteGenericInterface> extends FastifyRequest<T> {
     user: User
   }
 }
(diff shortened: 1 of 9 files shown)


src/api/plugins/auth.ts
```diff
@@ -26,24 +23,21 @@ async function authPlugin(app: FastifyInstance) {
...
       }
 
-      const { user, newToken } = authService.verifyUser(token)
+      const { user, newToken } = authService.verifyToken(token)
 
       if (newToken !== undefined) {
         reply.headers['x-auth-token'] = newToken
       } else {
-        request.log.error('No user found')
+        request.log.error('Could not create new token for user', user)
```
moritzzingg-netlight, line 36: newToken can have the value undefined if the current token is still valid for longer than 15min, in that case it is not necessary to provide a new token. So catching the case that the token is undefined as an error is wrong here.

src/api/routes/auth.ts
```diff
@@ -1,40 +1,62 @@
...
+      try {
+          const token = await authService.authorizeUser(request.body.code);
+          reply.status(200).send({token});
+      } catch(error) {
+          console.log(error);
+          return reply.status(401).send()
+      }
+    }
+  )
+  app.post(
```
moritzzingg-netlight, line 35: Great work🚀

src/api/services/authService.ts
```diff
@@ -2,8 +2,9 @@ import axios from "axios";
 import apiConfig from "../../config/api"
 import { prisma } from "../../lib/prisma";
 import jwt from "jsonwebtoken";
+import { serviceAuthenticationBody } from "../types";
 
-interface Userinfo {
+interface GithubUserinfo {
```
moritzzingg-netlight, line 7: Not happy with this name as the user info acts both for github users and non-github clients, so keeping the more general name userinfo might be better, also that one is not perfect. But, to fully fix this we would have to separate real users and clients.

src/lib/api.ts
```diff
@@ -1,6 +1,23 @@
...
+      'Content-Type': 'application/json',
+    },
+    body: JSON.stringify({
+      client_id: 'garbo',
+      client_secret: secret,
+    }),
+  })
+
+  const token = (await response.json()).token
+  console.log(token)
```
moritzzingg-netlight, line 18: can be removed before merge

moritzzingg-netlight (changes requested):
Create work, just small things which I thing should be adjusted.

src/api/plugins/auth.ts
willeasp: aaaaah good catch

src/api/services/authService.ts
willeasp: On the phone now so can't double check but i think this interface is only used for interacting with the Github API

moritzzingg-netlight:
Regarding error messages, please review #647 to align with the suggested error messages in there

src/api/services/authService.ts
willeasp: This is only used for the github api so I think it makes sense to call it githubuser

irony (approved):
🚀 Snyggt jobbat!

Issue #633, opened 2025-02-18
[Feat] Update Wikipedia Page from Garbo


okis-netlight:
Subtask of #631 

## Description
Since we have data for scope 1, 2, and 3 emissions in the Garbo database, these can be used to update emissions data in Wikipedia articles.

## Tasks
To implement this we need to:
- Create a Wikipedia bot account
- Format Garbo emissions data to fit Wikipedias article format
- Update Wikipedia article using a bot account

PR #657, opened 2025-03-06
[#633] Update Wikipedia from Pipeline

okis-netlight:
Solves #633 

- Created Wikipedia library functions for updating articles and generating text
- Created a new worker in the Garbo pipeline that updates the wikipedia page based on an uploaded report


irony (approved):
🚀 Snyggt jobbat!

PR #704, opened 2025-03-26
Nullable email

EspenHegdal123:
making emails nullable in schema.prisma, AuthSErvices.ts and migration.sql

Making sure that the login for Github users is working for everyone on the team


EspenHegdal123:
Fixed in another PR

PR #705, opened 2025-03-26
Fix make email nullable

moritzzingg-netlight:
Co-Authored with @EspenHegdal123 

- Makes e-mail nullable to fix issue with users don't have set a public e-mail in github

### DIFF
--- a/prisma/schema.prisma
+++ b/prisma/schema.prisma
@@ -350,7 +350,7 @@ model Metadata {
 
 model User {
   id    String @id @default(cuid())
-  email String @unique
+  email String?
   name  String
   githubId   String? @unique
   githubImageUrl String?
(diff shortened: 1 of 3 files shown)


EspenHegdal123:
IT is done in another PR

prisma/schema.prisma
```diff
@@ -350,8 +350,9 @@ model Metadata {
 
 model User {
   id    String @id @default(cuid())
-  email String @unique
-  name  String
+  email String?
+  name  String @unique
+  displayName String?
```
willeasp, line 355: Maybe a personal preference or a bit nitpicky, but I feel like "name" and "displayName" have the same meaning. I think switching `name` to `username` will suffice

willeasp (changes requested):

PR #714, opened 2025-03-26
fix: do not hallucinate tables that arent there

irony:
We had problems sometimes when we try to extract tables from a PDF where there might be a big picture on the page:

image

Which lead to this result- because the prompt told very specifically to always return a table with n.a. if there was no relevant data. This confuses Garbo later on in the process.
image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,6 @@ prisma/generated
 *.db-journal
 *.pdf
 *.code-workspace
+
+# ignore secrets
+*secret*
--- a/src/prompts/followUp/scope12.ts
+++ b/src/prompts/followUp/scope12.ts
@@ -82,6 +82,21 @@ const queryTexts = [
   'Market-based and location-based emissions',
   'GHG protocol Scope 1 and 2 data',
   'carbon emissions CO2',
+  'CO2-eq',
+  'CO2 emissions',
+  'Scope 1 and 2',
+  'Scope 1',
+  'Scope 2',
+  'CO2e',
+  'tCO2e',
+  'GHG Protocol',
+  'emissions',
+  'carbon footprint',
+  'carbon accounting',
+  'carbon emissions',
+  'carbon emissions data',
+  'carbon emissions report',
+  'carbon emissions calculation',
 ]
 
 export default { prompt, schema, queryTexts }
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -13,3 +13,6 @@ prisma/generated
 *.db-journal
 *.pdf
 *.code-workspace
+
+# ignore secrets
+*secret*
```
irony, line 18: this is neccessary to ignore files that are there from another branch on my machine (feat/cdn) that has other yaml files for certs etc.

src/prompts/followUp/scope12.ts
```diff
@@ -82,6 +82,21 @@ const queryTexts = [
   'Market-based and location-based emissions',
   'GHG protocol Scope 1 and 2 data',
   'carbon emissions CO2',
+  'CO2-eq',
```
irony, line 85: here we add more search terms so we try to focus on the most relevant pages.

src/workers/nlmExtractTables.ts
```diff
@@ -66,6 +70,8 @@ const extractTextViaVisionAPI = async (
 
 const searchTerms = [
   'co2',
+  'co2e',
```
irony, line 73: quite important search terms that werent there..

src/workers/nlmExtractTables.ts
```diff
@@ -38,18 +38,22 @@ const extractTextViaVisionAPI = async (
...
       },
       {
         role: 'assistant',
         content:
-          'Sure. Sounds good. Send the screenhot and I will extract the table(s) and return in markdown format as accurately as possible without any other comment.',
+          'Sure. Sounds good. Send the screenhot and I will extract the table(s) if there are any and return in markdown format as accurately as possible without any other comment.',
       },
       {
-        role: 'assistant',
+        role: 'user',
```
irony, line 49: always a good practice to send data about context from the user- not the assistant.

src/workers/followUp.ts
```diff
@@ -22,7 +26,7 @@ const followUp = new DiscordWorker<FollowUpJob>(
       default: { schema, prompt, queryTexts },
     } = await import(resolve(import.meta.dirname, `../prompts/${type}`))
 
-    const markdown = await vectorDB.getRelevantMarkdown(url, queryTexts, 5)
+    const markdown = await vectorDB.getRelevantMarkdown(url, queryTexts, 15)
```
irony, line 29: We are also sending more pages - so the correct data could be included next time.

willeasp:
This was very helpful, thanks!

PR #738, opened 2025-04-09
add docling

willeasp:


moritzzingg-netlight:
Further development in  #756

PR #760, opened 2025-04-28
[635] Create new route and service for mailchimp

kaylawoodbury:
### Background
While testing the mailchimp integration from the FE, we found an issue for returning users trying to get a new download. The mailchimp form post only adds tags to new users, and does not for returning users, as the documentation seemed to indicate previously. So needed to set up a BE endpoint, which does allow for updating the tag on an existing user, thus triggering the rest of the automation process to send an email to the user. 

Eventually I want to bypass the mailchimp autmation completely and instead be able to have a single endpoint that will send the email + tag to mailchimp and start the download for the user, so we can both collect emails but bypass the sludge effect of making a user wait and access email first. But that can come as a pt 2.


src/api/services/mailchimp.ts
```diff
@@ -0,0 +1,71 @@
+import axios from 'axios'
+import crypto from 'node:crypto'
+
+const MAILCHIMP_API_KEY = process.env.MAILCHIMP_API_KEY!
```
moritzzingg-netlight, line 4: Would be a bit more safe to not directly access the environment variable, instead create a config file (you can copy the structure of wikipedia.ts)

src/api/services/mailchimp.ts
```diff
@@ -0,0 +1,71 @@
...
+const MAILCHIMP_SERVER_PREFIX = 'us14' // Change if needed
+const MAILCHIMP_LIST_ID = '133270842a'
```
moritzzingg-netlight, line 6: How variable are these values, if they change in the future, might be worth to also add them to the env variables

src/app.ts
```diff
@@ -112,12 +116,15 @@ async function publicContext(app: FastifyInstance) {
...
+  app.register(authentificationRoutes, { prefix: 'api/auth' })
   app.register(companyReadRoutes, { prefix: 'api/companies' })
   app.register(companyExportRoutes, { prefix: 'api/companies' })
   app.register(municipalityReadRoutes, { prefix: 'api/municipalities' })
   app.register(municipalityExportRoutes, { prefix: 'api/municipalities' })
-  app.register(companyPublicReportingPeriodsRoutes, { prefix: 'api/reporting-period' })
+  app.register(companyPublicReportingPeriodsRoutes, {
+    prefix: 'api/reporting-period',
+  })
+  app.register(downloadRequestRoute, { prefix: '' })
```
moritzzingg-netlight, line 127: add at least api prefix here and remove it in the endpoint definition

moritzzingg-netlight (changes requested):
One thing regarding the naming of endpoint and service function. I think I don't have the fully functional background, but having an endpoint called download-request calling a service function subscribeAndTagUser seems a bit odd. Either make the endpoint more general to subscribing to different things defined by the provided reason or not include the request parameter reason in the endpoint and statically define the reason in the endpoint instead.

src/api/routes/mailing-list.downloads.ts
```diff
@@ -1,20 +1,20 @@
...
+  app.post('/download-request', async (request, reply) => {
     const { email, reason } = request.body as { email: string; reason: string }
 
     if (!email || !reason) {
       return reply.status(400).send({ error: 'Email and reason are required' })
     }
 
     try {
-      await subscribeAndTagUser(email, reason)
+      await subscribeAndTagUser(email, reason, 'free-database-download-request')
```
moritzzingg-netlight, line 13: Yes, I guess this is the best way to keep the service general

moritzzingg-netlight (approved):
Looks good

PR #787, opened 2025-05-13
759 feat decouple pipeline from discord

### LINKED ISSUE #759: Decouple Pipeline from Discord
As we want to move away from Discord as the main UI interface for the Garbo pipeline, we have decouple the pipeline from Discord.
Currently, the jobs interact with the user through the created Discord thread. This causes issues when we want to enqueue jobs through the newly created validation tool.

The goal is to make the interaction with discord optional to allow to start and control jobs just through the validation tool while preserving legacy discord support.

Tasks:

- [ ] Make thread id optional
- [ ] Make message id optional (only set these when a thread id is present)
- [ ] Only send messages and prompts to discord if thread id is present

alexbinett-netlight:
Closes #759


EspenHegdal123 (approved):

src/lib/DiscordWorker.ts
```diff
@@ -102,10 +102,10 @@ function addCustomMethods(job: DiscordJob) {
   }
 
   job.setThreadName = async (name) => {
```
eli6, line 104: this seems like a workaround that isn't needed if we just do something like this in the precheck, if i understand correctly:

```
      if(job.data.threadId) {
        await job.setThreadName(companyName)
      }
```

i dont see a reason to call the setThreadName if there is no thread

src/lib/DiscordWorker.ts
```diff
@@ -26,7 +26,7 @@ export class DiscordJob extends Job {
   ) => Promise<
     OmitPartialGroupDMChannel<Message<true>> | Message<true> | undefined
   >
-  setThreadName: (name: string) => Promise<TextChannel>
```
eli6, line 29: if we just avoid calling this function when there is no thread, then we don't need the undefined return type here (see comment below)

src/lib/DiscordWorker.ts
```diff
@@ -57,14 +57,14 @@ function addCustomMethods(job: DiscordJob) {
   }
 
   job.sendMessage = async (msg: string) => {
-    message = await discord.sendMessage(job.data, msg)
+    message = job.data.threadId ? await discord.sendMessage(job.data.threadId, msg) : null
```
eli6, line 60: send message is called in so many places so i guess we need the workaround here for now :)

eli6 (commented):
take a look and see if you agree :) nothing major but there is one place where we dont need the workaround i think

PR #801, opened 2025-05-15
Split up followUp worker

### LINKED ISSUE #793: [REFACTORING] Split up the followUp worker
Currently, the followUp worker handles different tasks to ask the AI for for different types of data. As the worker is therefore executed at different points in the pipeline it makes it harder to follow the pipeline flow.

As a pipeline user, I want to better distinguish between the different followUp jobs to better understand the flow of the pipeline.

**Tasks**

- [ ] Summarize the common followUp worker logic into a abstract followUp worker
- [ ] For each different use case of the followUp worker create a new worker extending the abstract followUp worker
- [ ] Verify that the new workers function as expected

moritzzingg-netlight:
Closes #793

### DIFF
--- a/src/queues.ts
+++ b/src/queues.ts
@@ -17,7 +17,7 @@ export const QUEUE_NAMES = {
   FOLLOW_UP_FISCAL_YEAR: 'followUpFiscalYear',
   FOLLOW_UP_COMPANY_TAGS: 'followUpCompanyTags',
   FOLLOW_UP_BASE_YEAR: 'followUpBaseYear',
-  FOLLOW_UP_INDUSTRY_GICS: 'followUpIndustryGICS',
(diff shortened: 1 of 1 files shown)


src/lib/FollowUpWorker.ts
```diff
@@ -75,6 +78,23 @@ const followUp = new DiscordWorker<FollowUpJob>(
     job.log('Response: ' + response)
     return response
   }
-)
+  return job
+}
+export class FollowUpWorker<
+  T extends FollowUpJob
```
alexbinett-netlight, line 84: Add FollowUpType instead of T?

alexbinett-netlight (commented):
Here are small suggestions!

PR #806, opened 2025-05-21
Fix allow scope3 category value to be optional

moritzzingg-netlight:
That's currently preventing verifying scope 3 emissions without changing numbers


eli6 (approved):
looks great!

PR #830, opened 2025-05-28
[Fix] Add error handlnig for empty result from nlmParse and too many tokens in followUp prompt

moritzzingg-netlight:
- Added a docling fallback for the case that nlmParse returns nothing
- Added a error handling for the case that a prompt fails, because of too many errors.


EspenHegdal123 (approved):
Looks good!

eli6 (approved):
looks good to me too!

PR #850, opened 2025-06-05
[FEAT] Add company emission change calculation

### LINKED ISSUE #849: [Feature] Move company's emission change calculation into the API

moritzzingg-netlight:
Closes #849

### DIFF
--- a/src/api/services/companyService.ts
+++ b/src/api/services/companyService.ts
@@ -225,48 +225,87 @@ export function addCalculatedTotalEmissions(companies: any[]) {
 }
 
 export function addCompanyEmissionChange(companies: any[]) {
-  return( companies.map((company) => {
-    company.reportingPeriods.sort((a, b) => new Date(b.endDate).getTime() - new Date(a.endDate).getTime());
-    company.reportingPeriods.map((period: any, index: number) => {
-      if(index < company.reportingPeriods.length - 1) {
-        let adjustedCurrentTotal = 0;
-        let adjustedPreviousTotal = 0;
-        const previousPeriod = company.reportingPeriods[index + 1];
-        const { scope1: currentScope1, scope2: currentScope2, scope3: currentScope3 } = period.emissions || {};
-        const { scope1: preivousScope1, scope2: preivousScope2, scope3: preivousScope3 } = previousPeriod.emissions || {};
-        if(currentScope1 && preivousScope1) {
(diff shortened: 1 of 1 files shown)


src/api/services/companyService.ts
```diff
@@ -222,3 +223,50 @@ export function addCalculatedTotalEmissions(companies: any[]) {
...
+export function addCompanyEmissionChange(companies: any[]) {
+  return( companies.map((company) => {
+    company.reportingPeriods.sort((a, b) => new Date(b.endDate).getTime() - new Date(a.endDate).getTime());
+    company.reportingPeriods.map((period: any, index: number) => {
+      if(index < company.reportingPeriods.length - 1) {
+        let adjustedCurrentTotal = 0;
+        let adjustedPreviousTotal = 0;
+        const previousPeriod = company.reportingPeriods[index + 1];
+        const { scope1: currentScope1, scope2: currentScope2, scope3: currentScope3 } = period.emissions || {};
+        const { scope1: preivousScope1, scope2: preivousScope2, scope3: preivousScope3 } = previousPeriod.emissions || {};
```
hallski, line 236: There is a typo in `preivousScope1` here

src/api/services/companyService.ts
```diff
@@ -222,3 +223,50 @@ export function addCalculatedTotalEmissions(companies: any[]) {
...
+          adjusted: adjustedCurrentTotal > 0 ? ((adjustedCurrentTotal - adjustedPreviousTotal) / adjustedPreviousTotal * 100) : 0
+        }
+      } else {
+        period.emissionsTrend = {
+          absolute: null,
+          adjusted: null
+        }
+      }
+    });
+    return company;
```
hallski, line 270: Would be awesome to split this function up a bit to make it easier to read and test.

src/api/services/companyService.ts
```diff
@@ -222,3 +223,50 @@ export function addCalculatedTotalEmissions(companies: any[]) {
       }))
   )
 }
+
+export function addCompanyEmissionChange(companies: any[]) {
```
hallski, line 227: Do we have a type for the companies that we can use here to avoid disregarding type safety?

hallski (changes requested):
Overall I think having this come from the backend is a great addition and having both values allows us more options on how to present it in the UI.

eli6 (changes requested):

PR #855, opened 2025-06-11
Garbo emissions assessment

eli6:
this adds an api endpoint for creating an assessment by garbo of a company's emissions for one or several years. If there is more than one year Garbo will also analyse the change in emissions over those years.

it will later be a worker and part of the parsing flow but is a function behind an api now for testing it out.

There is a corresponding frontend commits that adds a button and a modal for calling this api.

### DIFF
--- a/src/prompts/followUp/emissionsAssessment.ts
+++ b/src/prompts/followUp/emissionsAssessment.ts
@@ -1,31 +1,6 @@
 import { z } from 'zod'
 
-// Modified schema to be compatible with OpenAI's validation
-export const schema = z.object({
-  assessment: z.object({
-    isReasonable: z.boolean(),
-    confidence: z.number(),
-    issues: z.array(z.object({
-      type: z.enum(['MISSING_DATA', 'CALCULATION_ERROR', 'SCOPE_MISSING', 'UNIT_ERROR', 'OTHER', 'UNREASONABLE_REDUCTION']),
-      description: z.string(),
-      severity: z.enum(['LOW', 'MEDIUM', 'HIGH']),
-      suggestedAction: z.string().optional(),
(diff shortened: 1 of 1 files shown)


src/api/routes/emissionsAssessment.ts
```diff
@@ -0,0 +1,114 @@
...
+import { companyWithEmissionsInclude } from '@/jobs/emissions/types'
+import { assessmentResultSchema, errorResponseSchema } from '@/jobs/emissions/schema'
+import { assessEmissions } from '@/jobs/emissions/assess'
+
+const emissionsAssessmentSchema = z.object({
+  wikidataId: z.string(),
+  years: z.array(z.string())
+})
+
+const getCompanyEmissionsDataFromDatabase = async (wikidataId: string, years: string[]): Promise<CompanyWithEmissions | null> => {
```
irony, line 14: I think we already have this method somewhere?

src/prompts/followUp/emissionsAssessment.ts
```diff
@@ -0,0 +1,88 @@
+import { z } from 'zod'
+
+// Modified schema to be compatible with OpenAI's validation
+export const schema = z.object({
+  assessment: z.object({
+    isReasonable: z.boolean(),
+    confidence: z.number(),
+    issues: z.array(z.object({
```
irony, line 8: shouldnt we reuse the issue schema here?

irony (approved):

src/api/routes/emissionsAssessment.ts
eli6: we have none that filters by years so i think it can stay for now, but I'll keep it in mind!

Issue #857, opened 2025-06-11
[FEATURE] Improve the staging-prod evaluation script


alexbinett-netlight:
We want to improve the metrics used to calculate the Garbo accuracy. Also, we want a better overview of which reports contains errors, and potentially what type of errors.

kaylawoodbury:
Closing this issue as it seems the original intent was done, we'll likely want to come back to this evaluation script in other capacities over time

PR #859, opened 2025-06-13
Add correct calculation of accuracy, and add "error-identifying" metrics

alexbinett-netlight:
We want to improve the metrics used to calculate the Garbo accuracy. Also, we want a better overview of which reports contains errors, and potentially what type of errors.

Connected to: #857


alexbinett-netlight:
**Current interpretation of accuracy:**
Correct values:
Values where the staging value is equal to the corresponding value in prod, 
values where both the staging value and the corresponding prod value are undefined (i.e. the report does not have a value).
Incorrect values: 
Values where the staging value is not equal to the corresponding value in prod.

scripts/utils/statisticsFunctions.ts
```diff
@@ -0,0 +1,127 @@
...
+  }
+
+  const numbHasActualValueAndIsExtracted = diffs.reduce((acc: number, current: Diff) => { return (current.productionValue !== undefined && current.stagingValue !== undefined) ? acc + 1 : acc;}, 0);
+  const numbExtractedValues = diffs.reduce((acc: number, current: Diff) => { return current.stagingValue !== undefined ? acc + 1 : acc;}, 0);
+  const numbActualValues = diffs.reduce((acc: number, current: Diff) => { return current.productionValue !== undefined ? acc + 1 : acc;}, 0);
+
+  // Out of all the extracted values that are actually a value, how many of them are correct and how many are not?
+
+  // Out of all the extracted values by Garbo (staging), how many are actually a value and how many are made up by Garbo?
+  const precision = {
```
eli6, line 32: what we are measuring here in precision is simply how often stage has a value when prod does not, which can be interesting to know i guess. i would call it 'hallucination rate' maybe rather than precision

for actual precision there is a formula that looks like this: `precision = correct_values / (correct_values + invented_values)` which we should use if we want to call it precision...   if we want the precision case we should also check how many values are correct and identical across prod+stage (`correct_values`) and add the invented values that are only in stage as `invented_values` (now we dont check for correctness at all here)

scripts/utils/statisticsFunctions.ts
```diff
@@ -0,0 +1,127 @@
...
+    description: 'Out of all the extracted values by Garbo, how many are actually a value and how many are made up by Garbo?',
+    value: numbHasActualValueAndIsExtracted/numbExtractedValues,
+    numbHasActualValueAndIsExtracted,
+    numbExtractedValues
+  }
+
+  // Out of all the true values (prod), how many of them could Garbo extract a value for and how many could it not?
+  const recall = {
+    description: 'Out of all the true values, how many of them could Garbo extract a value for and how many could it not?',
```
eli6, line 41: same with recall, here we are checking how often staging attempted to extract "something" where we actually expect "something". but we are not checking if it is correct which true recall would check..

recall:
`correct_values / correct_values + missed_values`

scripts/utils/statisticsFunctions.ts
```diff
@@ -0,0 +1,127 @@
...
+    value: numbHasActualValueAndIsExtracted/numbActualValues,
+    numbHasActualValueAndIsExtracted,
+    numbActualValues
+  }
+
+  // Out of all fields that are supposed to have a numerical value, how many are incorrect because of a magnitude error?
+  const magErr = diffs.reduce((acc: number, current: Diff) => {
+    return (current.productionValue !== undefined) &&
```
eli6, line 49: it would be good reuse the comparison logic from 'number of correct values' here in a function to avoid repeating conditions where we might have accidental typing errors

eli6 (changes requested):
quite dense this PR, but I made some initial comments for you to consider :)

scripts/utils/statisticsFunctions.ts
alexbinett-netlight: Yeah, we can rename it if you want! Just to be clear, my idea was to use precision and recall to measure Garbo's ability to extract values, not the values' correctness. To measure the correctness, I was thinking that we use our accuracy metric (the one without undefined).

**Sidetrack,**:
I defined the positive class as 'the existence of a value' and the negative class as 'the nonexistence of a value'. This was done because precision and recall is useful when the negative class is not as important as the positive class (since True Negatives are never used in the metrics). And since we want to ignore the scenarios where staging and prod both have no value, we define the negative class as 'nonexistence of value' which makes these scenarios the True Negatives that are ignored in the metrics.
In your example of the actual precision, I don't see how that is the actual formula. I struggle to understand what the positive and negative class would be defined as. To me, it seems like a mix of accuracy and precision.

eli6: yeah i dont think the name is important, we can keep measuring if garbo manages to extract values at all and call it maybe coverage or extraction rate or something. i dont think we need to overthink this either, the important thing is get some numbers that we know how to work with so we can improve the pipeline, but it is a bit fun to break down.. 

precision has a specific meaning that includes correctness so that confused me a little bit. 😅 

in reality i think we can have precision mesurements of different kinds, depending on what we define as the 'class'. so we can have precision for all non-undefined values, or precision where we define 'correct' as 'garbo has any value' if we want...

on a very high level as I see it:

**accuracy**: what we started out with, just a measure of correctness that includes everything, including undefined. not a very useful metric!

**precision**: we narrow it down and just check correctness on a subset of values that are relevant to us (quality of the values we actually get). there can be different precision metrics and the most global one we already get when we remove the undefined values.

**recall**:  basically a measure of if we found everything, comparing our correct values to all values that should have been there including missed ones.


so interestingly i think:

- If we check the number of correct values compared to all cases where stage has a value → 👉 Precision
- If we check correct prod values compared to all values that ideally should have been there (including ones missing from stage) → 👉 Recall

scripts/utils/statisticsFunctions.ts
alexbinett-netlight: Do you mean the comparison logic in the reduce of 'numbCorrectNumericalFields'? Because that is different from the comparison logic in this reduce, so I'm not sure I follow

scripts/utils/statisticsFunctions.ts
alexbinett-netlight: Yes, we could have different precision metrics based on how we define the classes, but not sure that is something we want to do at the moment though.

Yeah I agree with your high level explanations of accuracy, precision and recall.

I think it is important that we clearly define the classes we've decided on when implementing precision and recall, in order to make it super clear what the metric actually tells us. In the current precision we have 'existence of value' and 'nonexistence of value', which I believe is clear. If we, for example, want to limit the precision and only look at all the non-undefined values, we have to decide on another classification we find useful. The starting point when creating or changing precision and recall metrics should always be to decide on what the binary classes are.

Also, something I noticed during this discussion, we have to be careful when using the word 'correctness' as it has different meanings depending on the context and what we are measuring. As an example, with your description of accuracy, a field is correct if the values are equal (including when both values are undefined). But in the context of precision and recall, it is entirely dependent on what you have defined your classes as.

PR #877, opened 2025-06-25
Documenting accuracy

EspenHegdal123:
Following the same set up as the documentation as the Pipeline for Garbo has. This documentation will be continuously updated while we do modifications for the way of calculating the accuracy.


EspenHegdal123:
Add some overview comments of why we are calculating the accuracy like we do. Add this to the documentation.

PR #856, opened 2025-06-11
[Feat] Add generic function to ask the user for feedback

moritzzingg-netlight:
Closes #792

PR #880, opened 2025-06-29
Test pr 856

eli6:
This incorporates the work from PR #856 by @moritzzingg-netlight with merge conflicts resolved.

Original PR: #856
Changes include: splitting up the follow up workers and making each follow up worker individually responsible for handling user approvals instead of having that centrally in the saveToAPI file

⚠️ Please review carefully - this had many conflicts that I resolved manually.


src/workers/saveToAPI.ts
```diff
@@ -9,128 +7,71 @@ import { QUEUE_NAMES } from '../queues'
...
-          return { success: true }
-        } catch (apiError) {
-          const errorMessage = `Failed to save data to API: ${apiError.message || 'Unknown error'}`;
-          job.log(errorMessage);
-          console.error(errorMessage, apiError);
-
-          await job.sendMessage({
-            content: `❌ Error: Failed to save ${apiSubEndpoint} data for ${companyName}. Server returned an error: ${apiError.message || 'Unknown error'}`
-          });
+      await apiFetch(`/companies/${wikidataId}/${apiSubEndpoint}`, {
```
alexbinett-netlight, line 49: Pretty sure this will cause an error when apiSubEndpoint is an empty string. The string will be "/companies/${wikidataId}/" but the correct string should be without the "/" at the end. You can see how it was handled before on line 83 above this comment.

alexbinett-netlight (commented):
Just some small comments!

Maybe we should add even more of the code into the DiffWorker, as there was some repeating code between the different diffWorkers (diffBaseYear, diffGoals, etc).

PR #922, opened 2025-09-04
[921] update municipality climate data json

elvbom:


eli6 (approved):

elvbom:
> i think it looks great and can be merged! The only question that pops to mind is why the climate data objects each are in their own array?

Had an idea about this simplifying batch updates in the future, but didn't have time to explore it more so I'll remove it for now

PR #937, opened 2025-09-15
add major version as key for cache

elvbom:


eli6 (commented):
just a comment for now :)

PR #948, opened 2025-09-19
Create better defaults in .env.example

hallski:
The JWT_SECRET can not be empty or Garbo will through an error trying to login the user for local development.

The GITHUB_REDIRECT_URI was pointing to the wrong port which was really hard to tell since it was just two numbers swapped with each other.


elvbom (approved):
:) !

PR #958, opened 2025-09-29
[Fix] company caching

elvbom:
Base company cache on fingerprint made up of all company data instead of only metadata updates. Update prompted by former cache strategy breaking when introducing future emissions trend slopes on backend.

Also remove 304 response - this might not be want we want, then feel free to leave other suggestions we can commit!


elvbom:
> @elvbom I talked with @eli6 and I think we should be good to merge this, do you have any objections from your side?

Nope, go ahead!

PR #959, opened 2025-09-30
Latest Wikidata Upload Script

Melanomaniac:
This PR adds the updated "Push-to-wikidata" Upload script, which can be run manually to sync all of the company data from Klimatkollen's Side to Wikidata.

Additionally there are unit tests for some of the functions to make sure everything is working as expected! 👍


eli6:
The requirement we had from Wiki is that we should only sum categories and use that as total if there is no total already stated. If the company has stated a total value for Scope 3 we should use that value for Scope 3 and not sum categories. The stated total is available next to the Scope 3 categories in our data:

```
export interface Scope3 {
    categories: Scope3Category[],    
    statedTotalEmissions?: StatedTotalEmission
}

```

I can't see any total scope 3 value in the script right now, we just sum categories again which was the original version of the script.. maybe you can check with @EspenHegdal123  if this really is the version of the script where he fixed that, I think he did?

eli6 (changes requested):
see other comment :)

Melanomaniac:
This bug totally slipped past me, hopefully won't happen again >:(
The total carbon emission calculation should now be fixed as part of [0d04254](https://github.com/Klimatbyran/garbo/pull/959/commits/0d04254c8d2279ff99d532808984051bbd69d5a2)!
I've tested it multiple times with --dry-run and it seems to be calculating accurately now!

eli6 (changes requested):
I saw there was a re-run this morning, it looks correct now except the grand total where we should use the stated scope 3 value :). We currently sum the categories for the grand total (and explicilty skip any stated total), in this piece of code that must be changed to use the scope 2 total if it is available::


```
function calculateTotalEmissions(claims: Claim[]): number {
  return claims.reduce((total: number, current) => {
    if (!current.scope) return total

    // Skip certain claims based on conditions
    if (shouldSkipClaimForTotal(current, claims)) {
      return total
    }

    return total + parseClaimValue(current.value)
  }, 0)
}
```

For example, for the company EQT partners , 23,266 is the right grand total value (sum of all scopes using the stated scope 3 total), not 11 million which we get when we sum the categories:
Screenshot 2025-10-02 at 14 39 16

eli6:
Great addition with the dry run by the way 👏 !

Melanomaniac:
just ran the script one more time right now 👍

PR #980, opened 2025-10-15
[Fix] preserve report urls from other years

kaylawoodbury:
Update the logic on the reportUrl to preserve previous years url on only set for the current year in the report

### DIFF
--- a/src/workers/diffReportingPeriods.ts
+++ b/src/workers/diffReportingPeriods.ts
@@ -32,11 +32,16 @@ const diffReportingPeriods = new DiffWorker<DiffReportingPeriodsJob>(
       economy = [],
     } = job.data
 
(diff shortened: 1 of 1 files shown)


eli6 (approved):
looks good :) it seems the newest report isnt always the one accessible through the link, did it use to put the latest one we ran there and not necessarily the latest?

kaylawoodbury:
> looks good :) it seems the newest report isnt always the one accessible through the link, did it use to put the latest one we ran there and not necessarily the latest?

I'm trying to remember the previous functionality, ideally it should apply to the year of the report, and I guess not specifically to the latest year in the db. I was thinking that the logic would be grabbing the latest year that is noted in the report, as that should in theory be the reports reporting year.... if that makes sense 😅

PR #981, opened 2025-10-15
[FIX] Add company identifier object and api endpoints

kaylawoodbury:
Modifying how we store the lei number so we can make use of metadata and verified fields. Also to allow for possibility of storing other types of identifiers, none applicable atm but now there can be a structure there to support it.

This is also largely to enable better support for updating the lei numbers in editor and tracking validation.

This keeps the legacy lei field for now, similar to the path we used for the description field changes to avoid breaking anything.

### DIFF
--- a/prisma/schema.prisma
+++ b/prisma/schema.prisma
@@ -24,6 +24,9 @@ model Company {
   url         String?
   lei         String?
 
+  /// Company identifiers (LEI, Swedish org number, etc.)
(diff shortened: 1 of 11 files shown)


elvbom:
@kaylawoodbury is this still relevant?

kaylawoodbury:
Closing this as there is a new PR that implements this slightly differently

PR #994, opened 2025-10-20
[977] Added coatOfArmsUrl property to schema

### LINKED ISSUE #977: Add fetch for municipality logos from Wikidata.
Do display the municipality logos in the FE, we'd want to implement a fetch of the logos from wikidata in the BE and add to the api-response.

CarlWiren90:
To be able to display coat of arms on the FE, this PR adds the coatOfArmsUrl property to response schema as well as updating the climate-data.json to updated file including the coatOfArmsUrl property that is added via this PR -> https://github.com/Klimatbyran/municipality-data-pipeline/pull/53.

Closes #977

### DIFF
--- a/src/data/climate-data.json
+++ b/src/data/climate-data.json
@@ -1,7 +1,7 @@
 [
   {
     "name": "Ale",
-    "coatOfArmsUrl": "https://commons.wikimedia.org/wiki/Special:FilePath/Ale%20vapen.svg",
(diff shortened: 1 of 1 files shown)


kaylawoodbury (commented):
Hmm, I'm thinking about the FE implementation where we have the same components for the list view now, it might be simpler down the road if we use the same field name for both companies and municipalities, maybe like logo or image or something? Not a big deal, very manageable either way, just a thought

PR #1007, opened 2025-11-04
[fix] use scope1and2 and statedtotal scope 3 as fallback for calc total

kaylawoodbury:
Edit calculated total to use the combined scope1and2 value

### DIFF
--- a/src/lib/company-emissions/companyEmissionsCalculator.ts
+++ b/src/lib/company-emissions/companyEmissionsCalculator.ts
@@ -1,3 +1,90 @@
+import { Emissions } from '../emissions'
+
+/**
+ * Priority: mb > lb > unknown
+ * Returns null if no data is available.
+ */
+export function calculateScope2Total(scope2: any): number | null {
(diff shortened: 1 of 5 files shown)


src/lib/company-emissions/companyEmissionsCalculator.ts
```diff
@@ -41,25 +53,41 @@ function calculateEmissionTotals(currentPeriod: any, previousPeriod: any) {
...
     scope3: previousScope3,
+    scope1And2: previousScope1And2,
   } = previousPeriod.emissions || {}
 
-  // Compare Scope 1 emissions
-  if (currentScope1 && previousScope1) {
-    adjustedCurrentTotal += currentScope1?.total ?? 0
-    adjustedPreviousTotal += previousScope1?.total ?? 0
-  }
+  // Calculate scope 1+2 totals for each period independently
```
eli6: I would love it if we can make a small internal function `calculatePeriodTotal(scope1, scope2)` that we can call for both the current and prevoius periods below to get the current/previous Scope1And2Total, to avoid repeating the code for both periods here and make it more readable  :)

PR #1014, opened 2025-11-10
[Fix] Added logoUrl to prisma.schema & changed company.read return

CarlWiren90:
Added logoUrl to priscma.schema and changed the company.read return to not return cache.


kaylawoodbury:
Note for future: this is a test PR, will need to revert to using cache after test

PR #1064, opened 2026-01-15
[1063] update territorial data for 2026

### LINKED ISSUE #1063: Update territorial data
Update municipality data to include

- updated data for 2026 (our own calculations, no new data from SMHI) - approximated, trend, total trend and total carbon law
- add hit net zero date
- include emissions from cement
- add emissions slope

Update regional data to include
- coat of arms
- add hit net zero date
- sector emissions

Update national data
- first commit

elvbom:
Update territorial data for 2026 as well as

Municipalities:
- remove deduction of emissions from cement

Regions:
- add coat of arms
- add political rulse and RSO

Add national data.

Closes #1063

### DIFF
--- a/src/api/routes/national.read.ts
+++ b/src/api/routes/national.read.ts
@@ -11,7 +11,7 @@ import { redisCache } from '../..'
 import fs from 'fs'
 import apiConfig from '../../config/api'
 
-const NATION_CACHE_KEY = 'nation:data'
+const NATION_CACHE_KEY = 'nation:all'
 const NATION_TIMESTAMP_KEY = 'nation:timestamp'
 
 const getDataFileTimestamp = (filePath: string): number => {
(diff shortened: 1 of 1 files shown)


src/api/routes/national.read.ts
```diff
@@ -0,0 +1,97 @@
...
+      },
+    },
+    async (_request, reply) => {
+      const currentTimestamp = getDataFileTimestamp(apiConfig.nationDataPath)
+      const etagValue = `"${currentTimestamp}"`
+
+      const cachedNation = await redisCache.get(NATION_CACHE_KEY)
+
+      if (cachedNation) {
+        return reply.header('ETag', etagValue).send(cachedNation)
```
kaylawoodbury: I might be reading this wrong, but is the Etag here being set to the current file timestamp? Wondering if it should be set to the timestamp when the file was cached, or compare it the current file timestamp with the cached timestamp? We're storing NATION_TIMESTAMP_KEY and wondering if that should be used here 🤔... otherwise we can probably remove it

kaylawoodbury (approved):
Nice! 🚀 

I didn't look at all the data in the changed data json files, but the api changes seemed to look good overall. I have one question about the caching/Etag, but it's not my strong suit so it might not be an issue 🙂

PR #1082, opened 2026-02-02
[Fix] Improvements in calculating scope 3 total calculation and fallback to statedTotal

CarlWiren90:
This pull request updates the logic for calculating Scope 3 emissions totals to better handle edge cases, and adds new tests to ensure correct behavior. The main change is that the total now returns null if all category totals are null or undefined, allowing fallback to stated totals. Comprehensive tests have been added to verify these scenarios.

**Calculation logic improvements:**

* Updated `calculateScope3CategoriesTotal` in `companyEmissionsCalculator.ts` to return null if all category totals are null or undefined, instead of returning 0. This change ensures that callers can correctly fall back to `statedTotalEmissions` when categories carry no numeric values.

**Testing enhancements:**

* Added a new test suite for `calculateScope3Total` in `companyEmissionsCalculator.spec.ts`, covering cases where categories are missing, all totals are null/undefined, totals sum to zero, or neither categories nor stated totals exist.


kaylawoodbury (approved):
The commit history looks a little wonky 😅 but the changes themselves look great! ⭐ Nice work updating the logic and adding new tests 🙂

PR #1084, opened 2026-02-02
[1083] Added internal companies and municipality endpoints

CarlWiren90:
This pull request introduces new internal API routes for reading company and municipality data, while also refactoring and reorganizing existing public routes for better code structure and maintainability.

**Key changes:**

### New Internal API Routes

* Added `internalCompanyReadRoutes` in `src/api/routes/internal/internal.company.read.ts`, providing /api/internal-companies. 
* Added `internalMunicipalityReadRoutes` in `src/api/routes/internal/internal.municipality.read.ts`, providing /api/internal-municipalities.


kaylawoodbury:
@elvbom Great question! 😄 

The primary driver for this separation is less about things being irrelevant for external consumers but rather to have a curated and safe guarded endpoint for external clients. Atm whenever we want to change the endpoints, to test some new item or make changes mainly for our FE or validation tool, we need to give advanced notice to our clients about the change so they have time to adjust. This has delayed us previously with releases. 

By separating, we can move at any speed we want with internal endpoints but be more thoughtful, and proactive with dedicated changes to the endpoint external client use and ensure it's only relevant data for them. 

Long term the vision would be to have some form of versioning so that clients would have time to adapt at their own pace, but this is a fairly quick and simple initial approach to decouple internal work from impacting clients unintentionally.

elvbom:
> @elvbom Great question! 😄
> 
> The primary driver for this separation is less about things being irrelevant for external consumers but rather to have a curated and safe guarded endpoint for external clients. Atm whenever we want to change the endpoints, to test some new item or make changes mainly for our FE or validation tool, we need to give advanced notice to our clients about the change so they have time to adjust. This has delayed us previously with releases.
> 
> By separating, we can move at any speed we want with internal endpoints but be more thoughtful, and proactive with dedicated changes to the endpoint external client use and ensure it's only relevant data for them.
> 
> Long term the vision would be to have some form of versioning so that clients would have time to adapt at their own pace, but this is a fairly quick and simple initial approach to decouple internal work from impacting clients unintentionally.

Then I'm onboard! :)

PR #1085, opened 2026-02-04
Feat european data

elvbom:


elvbom:
Closing this since we're using the Climate TRACE API instead

PR #1093, opened 2026-02-13
[1091] Added endpoint for getting only company names

CarlWiren90:
Added endpoint for getting a list of all company names for starting testing with crawler.


src/api/schemas/response.ts
```diff
@@ -1,6 +1,8 @@
 import { z } from 'zod'
 import { extendZodWithOpenApi } from '@asteasolutions/zod-to-openapi'
 import { emissionUnitSchemaGarbo, wikidataIdSchema } from './common'
+import { Language } from '@prisma/client'
```
elvbom: Is this used anywhere?

PR #1101, opened 2026-02-17
[Fix] Update company delete to include metadata

kaylawoodbury:
Cascade company delete to also remove linked metadata


kaylawoodbury:
Pausing this PR until we can discuss as a group and confirm we want cascading deletes, atm this does impact all metadata. We could leave the metadata unconnected if we want to be able to go back to it for some reason. 

Will set up call for this on Thursday.

kaylawoodbury:
Will keep metadata for now on deleted data, may need to clean db later if needed.

PR #1154, opened 2026-03-19
[1153] create new archive table for the queue

### LINKED ISSUE #1153: Explore setting up a table for queue data
Rather than keeping bull mq data in redis longer term, lets store it in a db table in postgres

kaylawoodbury:


kaylawoodbury:
this ended up shifting a bit and is no longer relevant

PR #1182, opened 2026-04-14
[FEAT] initial set up for api key

kaylawoodbury:
## Overview

Adds an `X-API-Key` authentication gate to all client-facing routes under `/api/...`, replacing the previous unauthenticated access. Keys use the format `garb_<lookup>.<secret>` and are role-based. Validate/Bolt attach the key via a Vite proxy header (`GARBO_PROXY_CLIENT_API_KEY`).

### Auth gate
- New `clientApiKeyGate` Fastify plugin runs on every request to `/api/*`, resolving the required permission from a central registry and verifying the key against a hashed+peppered secret stored in the DB.
- `ALLOW_ANONYMOUS_CLIENT_API=true` bypasses the gate entirely — use this during cutover if needed.

### Permission registry
- `src/api/security/routePermissions.ts` — ordered rules mapping `(METHOD, path)` → permission code. First match wins. Covers all current client routes including the new global search endpoint added during rebase.

### Route restructure
- All client-facing routes moved from `publicContext` in `app.ts` into `registerClientApiRoutes.ts`, making the X-API-Key surface explicit.
- Internal-only admin routes stay in `authenticatedContext` (JWT required).

### Admin endpoints (JWT only)
New routes under `POST /api/internal/client-api-keys`:
- `GET /roles` — list roles and their permission codes
- `GET /` — list keys (no secrets returned)
- `POST /` — create key; plaintext returned once
- `POST /:id/revoke` — soft-delete (sets `revokedAt`; key stops working immediately)
- `GET /endpoint-catalog` — export the full permission registry (useful for Validate UI)

### Data model
New Prisma models: `ClientApiPermission`, `ClientApiRole`, `ClientApiRolePermission`, `ClientApiKey` (with `keyLookup`, `secretHash`, `roleId`, `revokedAt`, `lastUsedAt`).

Two migrations included:
- `20260413120000_client_api_keys` — core tables
- `20260512100000_client_api_key_last_used_at` — adds `last_used_at` column, updated non-blocking on each successful auth

### Roles and seeding
Two built-in roles seeded via `npm run seed:client-api` (safe to run in production without touching other data):
- `all_access` — full client API surface (for Validate/Bolt)
- `base` — companies list/read/search only (for external partners)

Seed keys can be set via `GARBO_ALL_ACCESS_API_KEY` / `GARBO_BASE_API_KEY` env vars.

### Other
- `src/lib/redisCacheSingleton.ts` — extracted `redisCache` out of `src/index.ts` to break a circular import introduced by the route restructure. Causes a wide but trivial one-line import change across route files.
- In-process rate limiting per key with a `setInterval` pruner (`.unref()`) to prevent unbounded memory growth.
- `jest.env-setup.cjs` updated with `REDIS_HOST`/`REDIS_PORT` defaults so tests don't require a running Redis.

## Testing

- `src/registerClientApiRoutes.registry.test.ts` — every `/api` route has a registry entry
- `src/api/security/routePermissions.test.ts` — registry consistency (no duplicate rules, all permission codes covered)
- `src/api/routes/internal/clientApiKeys.admin.test.ts` — revoke (404/409/200) and endpoint-catalog routes
- `src/lib/clientApiKeyCrypto.test.ts` — key parsing, hashing, timing-safe comparison
- `src/config/parseEnvBoolean.test.ts` — guards against `"false"` string coercion bug

## Deployment notes

- Run `npm run prisma migrate dev` then `npm run seed:client-api` before deploying.
- Set `GARBO_ALL_ACCESS_API_KEY` in the environment before seeding (or create via the generate key API endpoint after deploy).
- Set `GARBO_PROXY_CLIENT_API_KEY` in Validate/Bolt `.env` so their Vite proxies attach the key.
- Use `ALLOW_ANONYMOUS_CLIENT_API=true` temporarily if you need a zero-downtime cutover window.
- See `doc/API_KEYS.md` for full implementation detail and troubleshooting.


eli6:
Looks good from what i saw this far (but i focused mostly on high level setup and didnt dive into the code in detail)!

A few things:
- For the rate limiting, it is just for one process but we can scale to quite many workers (we have 2 normally) so it would make sense to put this in redis for it to work across pods. If we track usage like i suggest below we would know what 'normal' usage looks like which can be interesting for this.

- We just log API calls in the process now and they will be hard to analyse that way, but it would be great to have a small db table with api key usage from the clients, with timestamp and which endpoint they called (maybe not the `base` one from our frontend to not pollute our table). That way we can easily query how many times they called/used endpoint X. If they do massive amounts of calls we could batch this but we dont know yet how much they call us :D

Something like this:

```
ClientApiRequest
  keyId
  path
  method
  statusCode
  timestamp

```

- Is "base" meant to be used by our frontend now, and "all_access" is for partners @kaylawoodbury ?

kaylawoodbury:
Great idea! I've added in the usage table in this PR, but I'll do a follow up PR for the rate limiting related change 🙂 

To answer you question, it's the inverse. So base is for base level api usage for a limited number of endpoints (atm external company endpoints) and the all access is what we would use for our FE that supports all endpoints including internally facing ones.

PR #1190, opened 2026-04-15
[1186] rename jobs and workers when removing discord as interface

### LINKED ISSUE #1186: Rename jobs and workers related to Discord
Rename jobs and workers related to Discord

elvbom:
Rename Discord workers and jobs, since we're removing the possibility to run reports via Discord and the naming is misleading.

Also fix some warnings and errors along the way.

Closes #1186


src/lib/PipelineQueue.ts
```diff
@@ -1,16 +1,16 @@
-import { Queue, QueueOptions } from 'bullmq'
+import { JobsOptions, Queue, QueueOptions } from 'bullmq'
 import redis from '../config/redis'
 
-export type DiscordJobData = {
+export type PipelineJobData = {
   url: string
-  threadId: string
+  threadId?: string
   channelId?: string
```
eli6: Several of these, i think channeld, messageId (and possibly also threadId) are discord specific and that might not be as clear any more after the rename and will be harder to spot, so we could either also remove them and functionality related to them now, or leave a comment that they are a discord specific rest.

src/lib/PipelineWorker.ts
```diff
@@ -101,7 +107,7 @@ function addCustomMethods(job: DiscordJob) {
     )
   }
 
-  job.sendMessage = async (msg: string) => {
+  job.sendMessage = async (msg: string | BaseMessageOptions) => {
     if (!job.hasValidThreadId()) {
```
eli6: this threadId validation is also discord specific stuff! If we keep it now we can perhaps rename it to "discordThreadId" so we know what to remove in the next round

eli6 (approved):
This looks good! But renaming does hide the discord coupling a bit though. PipelineWorker sounds more general than it is since it still has Discord specific stuff. I think it can be good if we leave some comments highlighting what is still clearly discord related (thread validating, message id etc) so that we can easily identify it for removal in the next round :) I left a couple of comments.

elvbom:
Great feedback @eli6 , I'll fix!

elvbom:
@eli6 feel free to take another look! Updated some names to clarify Discord connection, as well as a few fixes for warnings i found along the way

PR #1210, opened 2026-04-17
[1208]: Added dedicated endpoint for global search functionality

### LINKED ISSUE #1208: [Change]: Create combinedSearch endpoint
To enchance and optimize when and how we fetch data, we should should start by having an endpoint for the combined search. At the moment we are doing three fetches (municipalities, companies, regions) on the FE and combining them there with a lot of datapoints that we do not need.

We should start by creating a dedicated combined search endpoint that can be used by the global search on the FE.

CarlWiren90:
This pull request introduces a new global search API endpoint that allows searching across companies, municipalities, and regions. The implementation includes new request and response schemas, a service for aggregating and ranking search results, and integration with the main application and OpenAPI documentation.

**Global Search API Implementation:**

* Added a new POST endpoint `/api/global-search` that enables searching for companies, municipalities, and regions in a single API call.
**Schema and Type Additions:**

* Introduced `globalSearchRequestSchema` and `globalSearchResponseSchema` for validating request and response payloads. (`src/api/schemas/request.ts`, `src/api/schemas/response.ts`)

**Integration and Documentation:**

* Registered the new route in the main Fastify app and included it in the public API context. (`src/app.ts`) 

Closes #1208


elvbom (approved):
Looking good! One though: should we require 3 chars instead of 1 for name? Thinking it might enhance user experience and prevent unnecessary API calls?

CarlWiren90:
> Looking good! One though: should we require 3 chars instead of 1 for name? Thinking it might enhance user experience and prevent unnecessary API calls?

So my thought here was that we debounce the input on the FE instead, since you'd might want to search for "AB" or just "A" to see what comes up? Wdyt?

PR #1217, opened 2026-04-23
[hotfix] coalesce company list cache rebuilds and add in-memory fallback when Redis is missing

elvbom:
This change reduces load and response times for GET company list routes when the Redis-backed cache is cold or Redis is temporarily unavailable.

- Single-flight loading: Concurrent requests for the same data fingerprint now share a single getAllCompaniesWithMetadata() load instead of each triggering a full rebuild.
- In-memory fallback: After a successful full load, the result is held in process memory (keyed by the same cache key as the Redis entry) so repeated requests are not forced back to the database when Redis reads fail.
- Refactor: The logic lives in a shared helper, getCompaniesListCached, used by both the public and internal “list all companies” routes.

Why Cold cache and Redis outages could cause many parallel requests to each run the full company list query and post-processing, which is expensive and can contribute to gateway timeouts. Coalescing and a small memory fallback address that class of issues without changing the API contract or response shape.

### DIFF
--- a/src/api/routes/external/company.read.ts
+++ b/src/api/routes/external/company.read.ts
@@ -6,6 +6,7 @@ import { getTags } from '../../../config/openapi'
 import { CompanySearchQuery, WikidataIdParams } from '../../types'
 import { cachePlugin } from '../../plugins/cache'
 import { companyService } from '../../services/companyService'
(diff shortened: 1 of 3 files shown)


elvbom:
Closing this since its growing a bit old for a hotfix

PR #1229, opened 2026-04-23
[FEAT] Added saving of redis data to new postgres tables

eli6:
## Summary

This branch adds a **read-only Postgres archive** of pipeline report runs: when workers observe BullMQ jobs completing or failing, Garbo upserts a **`ReportRun`** (per `threadId`) and appends a **`ReportRunJob`** row with queue metadata, optional prompt/query context, approval fields, and failure reasons. A small **authenticated HTTP API** exposes batches and runs for internal tools (e.g. Validate Jobbstatus Archive). A **one-off backfill script** can copy historical completed/failed jobs from Redis into the same tables.

Live queue behaviour (BullMQ) is unchanged except for the archival side effect and related wiring.

## Motivation

- Redis job history rolls off on a finite retention window; teams need **durable history** for debugging and reporting.
- **Batch** identity is normalized in Postgres: pipeline `job.data.batchId` may be a Garbo **`Batch.id`** (cuid) or a legacy human-readable string; both resolve to a stable **`Batch`** row linked from **`ReportRun`**.

## Database (Prisma)

New models:

| Model | Role |
|--------|------|
| **`Batch`** | Stable batch row: `id` (cuid), unique `batch_name` (same value as pipeline batch id / label). |
| **`ReportRun`** | One row per report thread (`threadId` unique): PDF URL, company/Wikidata, optional `batchDbId` → `Batch`, coarse `status`, timestamps. |
| **`ReportRunJob`** | One row per terminal BullMQ job (`completed` / `failed`): `jobId`, `queueName`, `wikidataId`, `approvedTimestamp`, `autoApprove`, `failedReason`, optional `prompt` / `queryTexts` / `markdown`, `startedAt` / `finishedAt`, FK to `ReportRun`. |

Migrations:

- **`20260429120000_report_run_queue_archive`** — Single squashed migration creating `Batch`, `ReportRun`, and `ReportRunJob` with the final column set (including `wikidataId`, `approved_timestamp`, `auto_approve` on jobs).

If a local database previously applied an **older** draft of this squash without those job columns, reset or fix the schema before relying on it (`prisma migrate reset` is simplest for dev-only DBs).

Before merging to `main`, ensure this branch includes **`main`’s latest migrations** so the migration timeline stays consistent in shared environments.

## HTTP API (`api/queue-archive`)

Mounted in `app.ts` with prefix **`api/queue-archive`**. Routes require the same auth as other internal Garbo APIs.

| Method | Path | Purpose |
|--------|------|---------|
| `GET` | `/batches` | List batches (`id`, `batchName`), optional `limit`. |
| `POST` | `/batches` | Body `{ "batchName" }`; upsert by name, returns `{ batch }` (for Validate creating a batch before enqueue). |
| `GET` | `/runs` | Paginated runs: `page`, `pageSize`, optional `q`, **`batchDbId`** (exact `Batch.id`), **`batchName`** (exact `Batch.batchName`). |
| `GET` | `/runs/:threadId` | One run with ordered jobs. |

Implementation: `src/api/routes/internal/queue.archive.read.ts` → `src/api/services/reportRunArchiveReadService.ts`.

## Workers (`startWorkers.ts`)

For each configured queue:

- **`QueueEvents`** listens for **`completed`** and **`failed`**.
- Loads the job, resolves **`ReportRun`** via `threadId` + PDF URL, resolves **`batchDbId`** via **`resolveReportBatchDbId`** (`src/lib/resolveReportBatchDbId.ts`).
- **`reportRunJob.create`** persists terminal state, metadata from `returnvalue.metadata` where present, and **`approvedTimestamp` / `autoApprove`** on success paths as appropriate.
- **`ReportRun.status`** is updated coarsely (e.g. failed on any failure; completed when the send-company-link step completes), aligned with existing pipeline semantics.

## Backfill

- **Script:** `scripts/backfill-report-runs-from-bullmq.ts`
- **npm script:** `npm run backfill:report-runs`
- Walks completed/failed jobs per queue (configurable `--dry-run`, `--queues=…`, `--chunk=…`), skips rows that already exist for the same `(queueName, jobId)`, then reconciles run status and run-level timestamps from persisted jobs.
- Env: `DATABASE_URL`, Redis settings per `src/config/redis.ts`.

## How to test

1. Run migrations against a dev database (`npm run migrate` or `prisma migrate dev`).
2. Start API + workers; run a small pipeline that completes (or fails) a step; confirm **`ReportRun`** / **`ReportRunJob`** rows in Postgres.
3. Call **`GET /api/queue-archive/runs`** (authenticated) and **`GET /api/queue-archive/runs/:threadId`** with a known `threadId`.
4. Optionally run **`npm run backfill:report-runs -- --dry-run`**, then without `--dry-run` on a staging Redis snapshot.

### 📋 Checklist

- [ ] PR title starts with `[#issue-number]`; if no issue is applicable use: `[fix]`, `[feat]`, `[ops]`, or `[prod]`
- [ ] I’ve added/updated tests (or noted why they’re not needed)

### 🛠 Related Issue

Closes #[issue-number]


scripts/backfill-report-runs-from-bullmq.ts
```diff
@@ -0,0 +1,412 @@
...
+  const startedAt =
+    typeof job.processedOn === 'number' ? new Date(job.processedOn) : null
+  const finishedAtMs =
+    typeof job.finishedOn === 'number'
+      ? job.finishedOn
+      : typeof job.processedOn === 'number'
+        ? job.processedOn
+        : typeof job.timestamp === 'number'
+          ? job.timestamp
+          : Date.now()
```
eli6: I don't think we want the current timestamp if we cant find any data for the time, better to have it as undefined then i think? maybe there is a way of finding out what the exact attribute for the time is so we dont need the fallbacks (just less risk of getting it wrong here, i know its a one off script ^^)

scripts/backfill-report-runs-from-bullmq.ts
```diff
@@ -0,0 +1,412 @@
...
+        ? (data as { autoApprove?: unknown }).autoApprove
+        : false
+    )
+
+    await prisma.reportRunJob.create({
+      data: {
+        jobId,
+        queueName,
+        status,
+        wikidataId: wikidataId ?? null,
```
eli6, line 223: we add the wikidata to the ReportRun so maybe we don't need to store it on each job, that might just be noise actually

eli6 (commented):
i find nothing off although the one-off script isn't super easily readable 😅 , i left two comments of things that stood out.

what matters more is that i think a port-forward wouldn't work here, the forwarding is a bit shaky and we would most likely fail when transmitting this much data that way, so i commited a Kubernetes job manifest for us that we can apply to our cluster :

`kubectl apply -f k8s/jobs/backfill-report-runs.yaml -n garbo-stage`

that will start up a garbo pod that only runs the npm backfill script and then stops running when it is done.

scripts/backfill-report-runs-from-bullmq.ts
kaylawoodbury:
```suggestion
          : null
```

PR #1251, opened 2026-05-05
Robustify Docling error handling

eli6:
## Summary

- Limit BullMQ concurrency to **max 1** across all worker instances — Docling handles one request at a time, previously allowed 2
- Retry initial POST submission on gateway errors (502/503/504) with jitter — covers cases where the container is temporarily unreachable or recovering
- Add **jitter** to exponential backoff to prevent thundering herd when multiple jobs retry simultaneously
- Extend initial POST timeout from 30s to 60s — Docling may be slow to accept when busy
- On **404 during status polling**, clear `taskId` from job data so the next BullMQ retry resubmits fresh instead of polling a dead task (happens when Docling container restarts and loses in-memory state)
- Retry `doclingParsePDF` jobs up to **3 times** with 30s backoff via BullMQ `attempts`
- Gateway errors (502/503/504) throw a retryable `Error`; all other failures throw `UnrecoverableError` so BullMQ does not retry them
- Docling-side task failure throws `UnrecoverableError` immediately — no point retrying
- `fetchWithRetry` internal retries now only apply to gateway errors, not all 5xx (e.g. 500 is not retried)

## Test plan

- [ ] Verify jobs queue correctly and only one hits Docling at a time
- [ ] Simulate Docling returning 502 during submission — confirm retry with backoff
- [ ] Simulate Docling returning 500 — confirm job fails immediately without retry
- [ ] Simulate Docling restart mid-job — confirm 404 clears taskId and next retry resubmits

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 node_modules
+local/
 dist
 .env
 .aider*
--- a/src/workers/doclingParsePDF.ts
+++ b/src/workers/doclingParsePDF.ts
@@ -430,7 +430,9 @@ const doclingParsePDF = new DiscordWorker(
         )
       }
     } catch (error) {
-      job.log('Error: ' + error)
+      job.log(
+        `[${new Date().toISOString()}] attempt ${job.attemptsMade}/${job.opts.attempts} failed: ${error}`
+      )
       job.editMessage(
         `Failed to parse PDF: ${error.message || 'Unknown error'}`
       )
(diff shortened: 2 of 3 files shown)


kaylawoodbury (commented):
Changes seem reasonable 🙂 Are you testing locally or mainly plan to in stage to simulate prod better?

eli6:
> Changes seem reasonable 🙂 Are you testing locally or mainly plan to in stage to simulate prod better?

I have tested all scenarios locally and will merge to stage now so we can test run there :) hopefully we will see less reports stuck on the docling steps since they might be able to auto-recover

PR #1252, opened 2026-05-05
[1235] Pt1 dedup reports table 

### LINKED ISSUE #1235: [Change Request]: Add new field to reporting period
### Background / Why This Change?

We need to start shifting the reporting periods to be stored by report id (many to 1 relationship)

### What Should Be Changed?

Add report id (from report table) to the reporting period table.

### Compatibility / Migration Notes (optional)

_No response_

### Urgency

Nice to have

kaylawoodbury:
## Problem

The `Report` table (registry) was creating duplicate rows for the same PDF document. The root cause was a mismatch in how different write paths stored the URL:

- **Crawler** saves the web link in `url` (e.g. `https://company.com/report-2024`)
- **Pipeline** (old behaviour) stored the GCS/S3 URL in `url` and the web link in `sourceUrl`

Because the old `upsertReportInRegistry` only looked up existing rows by `sha256 → sourceUrl → url` sequentially, a pipeline save with `url = s3Url` never matched the crawler's row with `url = webUrl`, so a second row was created for the same document.

### Core fix — multi-key lookup with cross-links (`src/api/services/`)

**New file: `registryReportIdentity.ts`** — pure helpers extracted from `registryService.ts`:
- `buildReportLookupOr` — builds a Prisma OR query across all four identity fields (`sha256`, `sourceUrl`, `url`, `s3Url`) plus two **cross-link clauses** that bridge the crawler/pipeline naming split:
  - `{ url: input.sourceUrl }` — finds a crawler row (keyed by web URL in `url`) when the pipeline sends that same URL in `sourceUrl`
  - `{ sourceUrl: input.url }` — reverse case
- `pickSurvivorReport`, `mergeNullReportFields` — survivor selection and null-coalescing for merge operations
- `isLikelyStoredObjectUrl` — heuristic to detect GCS/S3/CDN URLs
- `identityScore` — counts non-null identity fields for survivor selection

**Updated: `registryService.ts`**
- `upsertReportInRegistry` now uses `findMany` with `buildReportLookupOr` instead of sequential `findUnique` calls
- Handles three cases: 0 matches → create; 1 match → update; 2+ matches → transaction merge (existing duplicates cleaned up on next write)
- Renamed `promoteHumanUrlFromInput` → `upgradeToWebUrlIfAvailable` (clearer intent)

### Pipeline payload fix (`src/workers/saveToAPI.ts`)

`pickRegistryPayloadFromReportingPeriodsSave` now correctly routes fields:
- `url` = web/link URL when available, otherwise GCS URL (file upload)
- `sourceUrl` = always stored when a non-GCS HTTP URL is known, even if it equals `url` — this is what distinguishes "confirmed web link" from "GCS-only"
- `s3Url` = GCS URL when available
- `sha256` = from `pdfCache.sha256` or `chosen.reportSha256`

Previously `sourceUrl` was suppressed when it equalled `url`, which lost the signal that a web link was known.

### Schema — partial unique on `s3Url`

`prisma/migrations/20260504120000_report_s3url_partial_unique/migration.sql`:
```sql
CREATE UNIQUE INDEX "Report_s3Url_key" ON "Report"("s3Url") WHERE "s3Url" IS NOT NULL;
```
Prevents two `Report` rows sharing the same GCS URL. Multiple `NULL` values remain allowed.

### Deduplication script (`scripts/dedupe-report-registry.ts`)

One-off cleanup for duplicate rows created before this fix. Uses union-find to cluster rows that share any identity field (including the `url ↔ sourceUrl` cross-link). Survivor selection: most non-null identity fields wins; tie-break: has `sha256` → lexicographically smallest `id`. Losers are merged into the survivor in a transaction.

```bash
npx tsx scripts/dedupe-report-registry.ts --dry-run
npx tsx scripts/dedupe-report-registry.ts --emit-mapping=./report-dedupe-mapping.csv
```

Run against staging first. k8s job manifest: `k8s/jobs/report-registry-dedupe.yaml`.

### Documentation (`doc/REPORT_REGISTRY.md`)

New developer reference covering:
- URL field semantics (`url`, `sourceUrl`, `s3Url`) and the `url = sourceUrl ?? s3Url` rule
- All four active write paths: Crawler, Validate registry add, Pipeline link upload, Pipeline file upload
- Legacy row shapes still in production
- How `upsertReportInRegistry` deduplicates (OR query + merge)
- How `pickRegistryPayloadFromReportingPeriodsSave` routes fields
- Dedup script usage and survivor selection rules
- Key source files table

### Tests

| File | New tests |
|---|---|
| `tests/registryReportIdentity.test.ts` | `buildReportLookupOr` — cross-links, no-duplicate guard, empty input |
| `tests/registryService.test.ts` | Cross-link integration (crawler row found when pipeline sends URL in `sourceUrl`); multi-row merge transaction |
| `tests/saveToAPI.test.ts` | New file — 10 tests for `pickRegistryPayloadFromReportingPeriodsSave`: URL-only, S3-only, hash+S3, cross-column, `reportYear` derivation, fallback to max year |

### Test environment fix (`jest.env-setup.cjs`)

Added dummy Discord env vars (`DISCORD_TOKEN`, `DISCORD_APPLICATION_ID`, `DISCORD_SERVER_ID`, `DISCORD_CHANNEL_ID`) to the Jest defaults. With `ts-jest` + `useESM: true`, `config/discord.ts` can be evaluated before the `DiscordWorker` mock takes effect; without these defaults the suite failed with `Invalid initialization of Discord environment variables`.

## How to deploy

1. Merge this PR and deploy the image
2. Run the dedup script against staging first, then production:
   ```bash
   kubectl create -f k8s/jobs/report-registry-dedupe.yaml
   ```
   Check `--dry-run` output before the live run.

## What this does NOT change

- No changes to the public API response shape
- No changes to `RegistryEditModal` or other Validate UI (that is PR 0 / Validate track)
- `@@unique([companyId, year])` on `ReportingPeriod` is untouched — that is PR 2

### 📋 Checklist

- [x] I’ve set labels, linked issue, and milestone (if applicable)

### 🛠 Related Issue

Closes #[issue-number]


src/api/services/registryReportIdentity.ts
```diff
@@ -0,0 +1,146 @@
...
+  const s = raw.trim().toLowerCase()
+  return (
+    s.includes('amazonaws.com') ||
+    s.includes('digitaloceanspaces.com') ||
+    s.includes('storage.googleapis.com') ||
+    s.includes('.r2.') ||
+    s.includes('minio') ||
+    s.includes('blob.core.windows.net')
+  )
+}
```
eli6, line 23: we already know we store with `https://storage.googleapis.com` so there is no reason to have a heuristic to check many other providers, we can just check if googleapis is part of the url and remove most of the above lines!

eli6: the "Likely" in the name is also quite redudant, unless accidentally some companies also store at storage.googleapis.com, not sure how we handle that case 😅

src/api/services/registryReportIdentity.ts
```diff
@@ -0,0 +1,146 @@
...
+  if (!trimStr(target.s3Key) && trimStr(donor.s3Key))
+    patch.s3Key = trimStr(donor.s3Key)
+  if (!trimStr(target.s3Bucket) && trimStr(donor.s3Bucket))
+    patch.s3Bucket = trimStr(donor.s3Bucket)
+  if (!trimStr(target.companyName) && trimStr(donor.companyName))
+    patch.companyName = trimStr(donor.companyName)
+  if (!trimStr(target.wikidataId) && trimStr(donor.wikidataId))
+    patch.wikidataId = trimStr(donor.wikidataId)
+  if (!trimStr(target.reportYear) && trimStr(donor.reportYear))
+    patch.reportYear = trimStr(donor.reportYear)
```
eli6: the above would be more easily maintainable without repetition and with a loop that says what is going on (i think it is easier to understand with 'rowToKeep', like this:

function copyMissingFields(
  rowToKeep: RegistryReportIdentityRow,
  rowToDelete: RegistryReportIdentityRow
): Partial<RegistryReportIdentityRow> {
  const patch: Partial<RegistryReportIdentityRow> = {}

  const fields = ['sha256', 'sourceUrl', 's3Url', 's3Key', 's3Bucket', 'companyName', 'wikidataId', 'reportYear'] as const

  for (const field of fields) {
    if (!trimStr(rowToKeep[field]) && trimStr(rowToDelete[field])) {
      patch[field] = trimStr(rowToDelete[field])
    }
  }

  const keepUrl = trimStr(rowToKeep.url)
  const deleteUrl = trimStr(rowToDelete.url)
  if (deleteUrl) {
    if (/^https?:\/\//i.test(deleteUrl) && !isStorageUrl(deleteUrl) && (!keepUrl || isStorageUrl(keepUrl))) {
      patch.url = deleteUrl
    } else if (isStorageUrl(deleteUrl) && !trimStr(rowToKeep.s3Url)) {
      patch.s3Url = deleteUrl
    }
  }

  return patch
}

src/api/services/registryReportIdentity.ts
```diff
@@ -0,0 +1,146 @@
...
+  r: Pick<RegistryReportIdentityRow, 'sha256' | 's3Url' | 'sourceUrl' | 'url'>
+): number {
+  let n = 0
+  if (trimStr(r.sha256)) n++
+  if (trimStr(r.s3Url)) n++
+  if (trimStr(r.sourceUrl)) n++
+  if (trimStr(r.url)) n++
+  return n
+}
+
```
eli6, line 38: here we can also just loop over those fields and maybe rename so we know that the function does a bit more? 

```
function numberOfIdentityFieldsInRow(row: RegistryReportIdentityRow): number {
  return ['sha256', 's3Url', 'sourceUrl', 'url'].filter(
    (field) => trimStr(row[field])
  ).length
}
```

src/api/services/registryReportIdentity.ts
```diff
@@ -0,0 +1,146 @@
...
+/** Survivor = richest identity; tie-break: has sha256, then lexicographically smallest id (stable). */
+export function pickSurvivorReport(
+  rows: RegistryReportIdentityRow[]
```
eli6, line 42: here i would explain in the code why number of identity fields makes the report 'win' so we can follow what happens. and preferably rename the ha/hb/a/ds fields to something with semantic meaning

eli6: example of more readable imo:
```

/**
 * Picks which duplicate row to keep. Prefers the row with the most identity
 * fields filled in (url, sourceUrl, s3Url, sha256), since that is likely the
 * pipeline row which has the most complete data. Tie-breaks: prefer the row
 * that has a sha256 (means we have a verified cached copy), then fall back to
 * alphabetical id for a stable result.
 */
export function pickRowToKeep(
  rows: RegistryReportIdentityRow[]
): RegistryReportIdentityRow {
  if (rows.length === 0) {
    throw new Error('pickRowToKeep: empty rows')
  }
  if (rows.length === 1) return rows[0]
  return [...rows].sort((a, b) => {
    const scoreDiff = numberOfIdentityFieldsInRow(b) - numberOfIdentityFieldsInRow(a)
    if (scoreDiff !== 0) return scoreDiff
    const aHasSha256 = trimStr(a.sha256) ? 1 : 0
    const bHasSha256 = trimStr(b.sha256) ? 1 : 0
    if (aHasSha256 !== bHasSha256) return bHasSha256 - aHasSha256
    return a.id.localeCompare(b.id)
  })[0]
}

```

src/api/services/registryService.ts
```diff
@@ -5,6 +5,96 @@ import {
...
+  type RegistryReportIdentityRow,
+  trimStr,
+  isLikelyStoredObjectUrl,
+} from './registryReportIdentity'
+
+/**
+ * If the existing row's `url` is an S3/CDN URL but the incoming input carries a proper
+ * web URL, return the web URL so the caller can upgrade the `url` field.
+ * Returns undefined when no upgrade is warranted (keep the existing value).
+ */
```
eli6: this is clearer, we know who the 'caller' is (our other function) so we can explain the context in which this function is used:

/**
 * When we already have this report stored with a storage.googleapis.com URL,
 * and the incoming data has a proper web link, return the web link so we can
 * update the stored url to something more human-readable so users of our frontend get the more readable link.
 */

eli6 (changes requested):

src/api/services/registryReportIdentity.ts
kaylawoodbury: Hmm good point, I guess we know we store them in a particular folder....maybe it's worth to even include that incase a company also uses an s3 link for their report on their website?

src/api/services/registryReportIdentity.ts
kaylawoodbury: I updated the comment but tried to use a slightly more paired down version of the suggestion. Let me know if it's clear enough, otherwise happy to update the comment text again 😄

kaylawoodbury:
@eli6 I updated the files quite a bit in terms of naming, comments etc. Let me know what you think and if it's a bit easier and more intuitive to follow now 🙂

Issue #707, opened 2025-03-26
Automate approving wikidata ID by checking against prod


willeasp:
### Background
When processing reports through the pipeline, the user is always asked about if the wikidata ID is correct. For companies that we already have in our database, this is done through copying the wikidataID from the discord chat and checking if it exists as a company on klimatkollen.se/companies/<wikidataID>. If it doesn't exist, a manual search is done to find the company on our site, where its wikidataID is copied and pasted into garbo. This entire step could be automated in the pipeline, saving tons of time approving companies that already exist in our database. 

### Solution
Automate checking against prod if the wikidataID exists in our database. If not, search for the name or ask the user.

PR #1263, opened 2026-05-11
[#1169] feat(guessWikidata): add forceWikidataReview flag to opt out of auto-approve

### LINKED ISSUE #1169: Add Upload Option to Force WikiData review
How it works now: 

When we run a pdf report it always runs "Guess wikidata" to find a wikidata for that pdf, one of two things then happen:

That wikidata is new, we don't have it in our database 👉  Garbo waits for us to confirm.
That wikidata already exists 👉  Garbo automatically uses it for the report and starts parsing the rest of the info. We can only approve or reject the individual updates later.

Problem: 
"Guess wikidata" can mistakenly pick the wrong wiki ID. (maybe the group instead of the company, or confuses a company with another, or thinks ICA latvia is ICA Group). 

If we already have that wrong wikidata for another company in the database, Garbo cannot be stopped and will just use that wikidata id and overwrite or try to overwrite the previous data.

Proposed solution:
Create a "this is a new report" flag in the validation tool. If that flag is turned on, Garbo will stop even if "Guess Wikidata" produces an ID we already have, since for a new company that is suspicious, we don't expect to find that wikidata ID already...

We can still approve that its the same cmopany but get a chance to prevent automatic confusion a company with another.

andersabjorn:
## ✨ What's Changed?

This pull request adds an opt-in `forceWikidataReview` flag to the `guessWikidata` worker that lets the uploader bypass the production-database auto-approve shortcut and force a manual review instead.

The change has two parts:

1. Extended `GuessWikidataJob.data` with an optional `forceWikidataReview?: boolean`. The flag propagates automatically from the parent `parsePdf` job through `precheck`, because `precheck.ts` spreads `...baseData` into its child jobs — no extra plumbing is needed inside garbo.
2. Wrapped the existing prod-DB auto-approve block in `if (!job.data.forceWikidataReview) { ... }`. When the flag is set, the worker skips the prod lookup entirely and falls through to the existing manual-review path further down in the same file. Default behavior is unchanged when the flag is absent or `false`.

Auto-approve when a wikidata ID is found in the production database was introduced deliberately in #707. This PR does not remove that behavior — it only adds a per-job opt-out for the cases #1169 describes, where the uploader suspects "Guess Wikidata" matched the wrong entity (for example ICA Latvia vs ICA Group).

Verified locally that the existing auto-approve path still runs when the flag is absent, and that the manual-review path is taken when `forceWikidataReview: true` is set on the job data.

## 📋 Checklist

- [x] PR title starts with [#issue-number]; if no issue is applicable use: [fix], [feat], [prod], or [copy]
- [x] Verified default behavior is preserved when the flag is not set
- [ ] End-to-end test pending — requires the matching pipeline-api PR so the flag can actually be set from upload

## 🛠 Related Issue

Closes #1169

Related:
- #707 — introduced the auto-approve behavior that this PR makes opt-out-able.
- `Klimatbyran/pipeline-api#45` — matching upload-side flag; will be linked here once that PR is opened.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -115,6 +115,7 @@
       "integrity": "sha512-CGOfOJqWjg2qW/Mb6zNsDm+u5vFQ8DxXfbM09z69p5Z6+mE1ikP2jUXw+j42Pf1XTYED2Rni5f95npYeuwMDQA==",
       "dev": true,
       "license": "MIT",
+      "peer": true,
       "dependencies": {
         "@babel/code-frame": "^7.29.0",
         "@babel/generator": "^7.29.0",
@@ -630,6 +631,7 @@
       "resolved": "https://registry.npmjs.org/@bull-board/ui/-/ui-6.21.3.tgz",
       "integrity": "sha512-s/PLBJab8cnoQAGVqjQb0v4oGe0KgB4aQ5G5g93doxzXB/D+wkXNL9P9+zLWLldBJXE57jL4CR99ttDCIiyNHw==",
       "license": "MIT",
+      "peer": true,
       "dependencies": {
         "@bull-board/api": "6.21.3"
       }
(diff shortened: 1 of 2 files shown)


kaylawoodbury (commented):
Changes look good! Just need to run `npm audit fix` so all checks pass before we can merge this to stage 🙂

andersabjorn:
> Changes look good! Just need to run `npm audit fix` so all checks pass before we can merge this to stage 🙂

Awesome, i've done that now and pushed changes. :)

PR #1279, opened 2026-06-02
[Change]: Add internal Company.id with migrations, backfill, and API exposure

CarlWiren90:
### ✨ What’s Changed?

# Company internal `id` migration

Adds a stable internal `Company.id` (UUID) while keeping `wikidataId` as the primary key and public identifier for now.

## Schema outcome

| Field        | Role                                              |
| ------------ | ------------------------------------------------- |
| `id`         | Internal unique ID, required, `@default(uuid())`  |
| `wikidataId` | Primary key (`@id`), Wikidata / API slug          |

Foreign keys (`companyId`, `companyWikidataId`, etc.) still reference `wikidataId`. No primary-key switch in this rollout.

## Migration

Single migration: `20260602120000_add_company_internal_id`

1. Adds nullable `Company.id`
2. Backfills existing rows with `gen_random_uuid()::text` in SQL
3. Sets `id` NOT NULL
4. Creates unique index on `id`

No separate backfill script or Kubernetes job.

### 📋 Checklist

- [x] If this changes the API contract, I’ve called that out clearly (and considered backward compatibility)
- [x] I’ve considered observability (logs/metrics/traces) and added what’s needed
- [x] I’ve set labels, linked issue, and milestone (if applicable)


package.json
```diff
@@ -89,6 +91,7 @@
     "@types/jsonwebtoken": "^9.0.8",
     "@types/node": "^22.10.6",
     "concurrently": "^9.1.2",
+    "cuid": "^3.0.0",
```
kaylawoodbury: This might need to be in dependency and not in devdependency

kaylawoodbury: Also looks like this package is deprecated and potentially flagged via audit. Might need to consider another package

``` 
"node_modules/cuid": {
  "version": "3.0.0",
  "resolved": "https://registry.npmjs.org/cuid/-/cuid-3.0.0.tgz",
  "integrity": "sha512-...",
  "deprecated": "Cuid and other k-sortable and non-cryptographic ids (Ulid, ObjectId, KSUID, all UUIDs) are all insecure. Use @paralleldrive/cuid2 instead.",
  "dev": true,
  "license": "MIT"
}
```

scripts/backfill-company-id.ts
```diff
@@ -0,0 +1,140 @@
+/**
+ * One-off backfill: populate nullable `Company.id` for existing rows.
```
kaylawoodbury: You might want to set this up as a k8 job so you can run a kuubectl command from your terminal instead

prisma/schema.prisma
```diff
@@ -14,7 +14,7 @@ generator client {
 model Company {
   // // NOTE: Maybe just use an integer as the ID for companies, and use the unique hash instead.
   // // We want to separate the company id from the URL slug
-  // id          String  @id @default(cuid())
+  id           String  @unique @default(cuid())
   /// wikidataId is our unique identifier to link to companies, also used to export to or import from wikidata
```
kaylawoodbury: might be better for a later PR but we should look to see if we need to update the comments on this file

scripts/lib/create-prisma-cuid.ts
```diff
@@ -0,0 +1,49 @@
+import { randomBytes } from 'node:crypto'
+import { hostname } from 'node:os'
+
+/**
+ * CUID v1 compatible with Prisma `@default(cuid())`.
+ * Inlined to avoid deprecated npm `cuid@3` (cuid2 uses a different format).
+ */
```
eli6: cuid v1 seems to be deprecated so I think it's best to use cuid v2, i dont think it matters if the format is slightly different (it solves security issues with cuid v1 which is a good thing). then we can also generate the CUIDs with this library @paralleldrive/cuid2 and not have this custom implementation :)

scripts/lib/create-prisma-cuid.ts
```diff
@@ -0,0 +1,49 @@
...
+
+function pad(value: string, size: number): string {
+  while (value.length < size) value = `0${value}`
+  return value
+}
+
+function randomBlock(): string {
+  const lim = 2 ** 32 - 1
+  const n = Math.abs(randomBytes(4).readInt32BE() / lim)
+  return pad((n * discreteValues).toString(base), blockSize)
```
eli6: the number here after division with lim is a decimal number most likely.. so the CUIDs will contain decimals and dots (.) which is not a valid CUID v1 strictly speaking. But if we can use the CUID v2 library above it will solve these kinds of bugs for us!

prisma/migrations/20260602092036_require_company_internal_id/migration.sql
```diff
@@ -0,0 +1,8 @@
+/*
+  Warnings:
+
+  - Made the column `id` on table `Company` required. This step will fail if there are existing NULL values in that column.
+
+*/
+-- AlterTable
+ALTER TABLE "Company" ALTER COLUMN "id" SET NOT NULL;
```
eli6: this won't work before the backfill and migration will fail when they auto run on deploy since we won't have id's initially!

doc/COMPANY_ID_MIGRATION.md
```diff
@@ -0,0 +1,153 @@
+# Company internal `id` migration
+
```
eli6: i am not sure we need this document in Garbo long-term, we could put what is in this 'docs' in the PR description maybe? 😅

eli6 (changes requested):
Nice work overall with this PR :)

I wonder why we choose CUID instead of UUID? CUIDs are slightly faster but I think thats negligeable for us.

if we use UUID we dont need a backfill script at all but can assign them easily and use one migration script 😅 :

ALTER TABLE "Company" ADD COLUMN "id" TEXT;
UPDATE "Company" SET "id" = gen_random_uuid() WHERE "id" IS NULL;
ALTER TABLE "Company" ALTER COLUMN "id" SET NOT NULL;

prisma/migrations/20260602092036_require_company_internal_id/migration.sql
CarlWiren90: So the thought around this would be running the migrations, this one would fail but the adding of id prop would pass, we'd backfill and run again. But perhaps it's then better to just have the id migration, merge, and then create and run the not null migration after?

doc/COMPANY_ID_MIGRATION.md
CarlWiren90: Good call! :D

prisma/migrations/20260602092036_require_company_internal_id/migration.sql
eli6: yeah if i remembr right the pod will be blocked by this and fail its initialization, so garbo will be down until we backfill and we don't want that 😅

src/api/schemas/common.ts
```diff
@@ -5,6 +5,9 @@ extendZodWithOpenApi(z)
 
 export const wikidataIdSchema = z.string().regex(/Q\d+/)
 
+/** Garbo internal `Company.id` (CUID). Distinct from `wikidataId`. */
```
eli6: comment now outdated, would be good to fix if we want to keep that comment. dont think we need to mention the type in the comment, it is clear from the code line

eli6 (approved):
just found an outdated comment, otherwise it is good to go!

src/api/schemas/common.ts
CarlWiren90: Great, removed it!

PR #1287, opened 2026-06-05
[fix] align READMEs with current pipeline and test setup

### LINKED ISSUE #771: Implementing LEI (Legal Entity Identifier) functionality for companies

elvbom:
### ✨ What’s Changed?

Update setup and flow docs for Docling-based PDF parsing, correct test suite commands, and document the BullMQ report-runs backfill job.

### 🔍 How to Review / Test

Read readmes :)

### 📋 Checklist

- [x] If this changes the API contract, I’ve called that out clearly (and considered backward compatibility)
- [x] I’ve considered observability (logs/metrics/traces) and added what’s needed

Also closes #771


doc/pipeline.md
```diff
@@ -240,7 +254,7 @@ Indexes the parsed markdown into the VectorDB for downstream retrieval.
 }
 ```
 
-Additionally, to the three standard fields the job gets three more input properties. Depending if the job was directly called by [nlmParsePDF](#nlmparsepdf) the field `cachedMarkdown` contains the cached markdown extraction which could be directly retrived from the VectorDB and passed to the job. The field `companyName` is deprecated and not used anymore, as the job determines the company name itself. The fiele `type` is also not set and only use to define the name of the wikidata schema.
+Additionally, the job may receive `cachedMarkdown` when [parsePdf](#parsepdf) skips Docling because the report is already indexed in the VectorDB. The field `companyName` is only used when the user provides it manually after automatic extraction fails. The field `type` is not set by the pipeline and is only used to define the name of the wikidata schema.
```
eli6: Even if the report is indexed in the vectorDB we can choose to parse it again, that's the choice of whoever runs the report (that setting is called force-reindex) :) but it is true that the default is to not parse it again if it has been cached

doc/pipeline.md
eli6, line 350: runOnly is now mentioned below but I notice that it is not mentioned here. it has type `(FollowUpKey | 'all')[]`

eli6: only two are passed for explicity use by extract emissions:

class ExtractEmissionsJob extends DiscordJob {
  declare data: DiscordJob['data'] & {
    companyName: string
    runOnly?: (FollowUpKey | 'all')[]
  }
}

the others are just base data sent everywhere (threadId) etc.. not relevant to extract emissions specifically

k8s/jobs/README.md
```diff
@@ -47,3 +47,16 @@ Sets `ReportingPeriod.companyReportId` for all rows. One `CompanyReport` per com
    ```
 
 6. Watch logs: `kubectl logs -n garbo-stage job/link-periods-to-company-reports-<suffix> -f`
+
+## Backfill report runs from BullMQ
+
+Creates or updates `ReportRun` / `ReportRunJob` rows from historical BullMQ job data (useful after schema changes or if run history was missing).
+
```
eli6: this was a one off script i believe. i dont think we need to keep it in the repo but we can check with the team

README.md
```diff
@@ -12,69 +12,70 @@ We utilise an open source queue manager called BullMQ which relies on Redis. The
 
 ## Current Status
 
-Test the app in Discord channel #reports-to-check by using the command /pdf and Garbo will be answering with a parsed JSON. Or start the validation frontend and pipeline-api to use our custom UI.
+Test the app in Discord channel #reports-to-check by using the command `/pdfs` and Garbo will process the report through the pipeline. Or start the validation frontend and pipeline-api to use our custom UI.
```
eli6: we don't support discord any more so we can remove that instruction

README.md
```diff
@@ -103,9 +104,9 @@ npm i
 
 ### Starting the containers
 
-This project expects some containers running in the background to work properly. We use Postgres as our primary database, Redis for managing the queue system, ChromaDB for embeddings and the NLM ingestor for parsing PDF:s.
+This project expects some containers running in the background to work properly. We use Postgres as our primary database, Redis for BullMQ, and ChromaDB for embeddings. PDF parsing uses **Docling** (hosted Berget API by default via `DOCLING_URL` / `BERGET_AI_TOKEN` in `.env`; optionally uncomment the `docling` service in `docker-compose.yaml` for local docling-serve).
```
eli6: they should use local docling (and uncomment it in the docker compose before starting it), we don't use hosting via berget any more and since hosting keeps changing i would leave it out where we host our internal version

eli6 (changes requested):
Great with a readme update!! I noticed some things that need adjustment

elvbom:
@eli6 whenever you have time for another look!

eli6 (approved):
looks a lot better now! feel free to merge it in :)

PR #1344, opened 2026-07-06
Feat/restore validated scope1

kaylawoodbury:
### 📋 Checklist

- [ ] PR title starts with `[#issue-number]`; if no issue is applicable use: `[fix]`, `[feat]`, `[ops]`, or `[prod]`
- [ ] I’ve added/updated tests (or noted why they’re not needed)
- [ ] I’ve verified the change runs locally (and in Docker/k8s if relevant)
- [ ] If this includes a DB change, I’ve included a migration + rollout/rollback notes

### 🛠 Related Issue

Closes #[issue-number]


scripts/restore-reporting-periods.ts
```diff
@@ -0,0 +1,601 @@
...
+              if (srcEconomy) {
+                await restoreEconomy(tx, srcEconomy.id, newPeriod.id)
+              }
+
+              console.log(
+                `   ✅ ${period.year}: created with full emissions + economy + metadata`
+              )
+              stats.periods_created++
+            }
+          } catch (err) {
```
kaylawoodbury: This try/catch is inside the target.$transaction callback. I think in Postgres, any failed statement (FK violation, unique violation, etc.) aborts the entire transaction, so not sure that the try catch will work on a per company basis 🤔

scripts/restore-reporting-periods.ts
```diff
@@ -0,0 +1,601 @@
...
+
+            // Find or create Report using the backup URL
+            let report = await tx.report.findFirst({
+              where: { OR: [{ url: reportUrl }, { sourceUrl: reportUrl }] },
+              select: { id: true },
+            })
+            if (report) {
+              console.log(`   Report: already exists in prod (reusing)`)
+            } else {
+              report = await tx.report.create({
```
kaylawoodbury: It might be helpful if we already have it to include fields like: wikidataId, companyName, reportYear for the report. I think having at least company name is helpful otherwise we'll have hundreds of "null" rows in the report table and won't know what company they belong to until we rerun the report

scripts/restore-reporting-periods.ts
```diff
@@ -0,0 +1,601 @@
...
+                select: { id: true },
+              })
+              await restoreMeta(
+                tx,
+                'reportingPeriodId',
+                newPeriod.id,
+                await lastMeta('reportingPeriodId', period.id)
+              )
+
+              const newEmissions = await tx.emissions.create({
```
kaylawoodbury, line 539: This seems like it will always create an empty emissions row even when the source period has none, but the economy data seems to be conditional. Do we want them to be the same behaviour?

scripts/restore-reporting-periods.ts
```diff
@@ -0,0 +1,601 @@
...
+}
+
+async function main() {
+  console.log(
+    `\n🔍 Mode: ${DRY_RUN ? 'DRY RUN (pass --commit to write)' : '✅ COMMIT'}\n`
+  )
+
+  // Companies where recovered max year is 2024 (optionally filtered to one)
+  const recoveredCompanies = values.company
+    ? [{ companyId: values.company }]
```
kaylawoodbury, line 375: Does this mean if you use the --company flag that it bypasses the 2024 year guard check?

eli6: Yes it does! But i just use the --company for testing so it won't be used in the bulk restoration process

scripts/restore-reporting-periods.ts
eli6: i am using the transactions for the dry run too, so i like it that it rolls back all. But the try catch is completely unnecessary because of that so I will remove it :D

scripts/restore-reporting-periods.ts
eli6: good catch, will fix!

scripts/restore-reporting-periods.ts
eli6: added them in :)

PR #1395, opened 2026-08-24
Add company alternativeNames foundation

kaylawoodbury:
## Summary
- Adds `Company.alternativeNames` (`String[]`, default empty) with a Prisma migration
- Strengthens name matching via `companyNameMatchKey` (case, spaces, periods, punctuation, legal suffixes, diacritics); `normalizeCompanyNameForMatch` delegates to it
- Adds `mergeAlternativeNames` / `isFormattingVariantOfCompanyName` for safe storage dedupe without collapsing distinct entities (e.g. Volvo AB vs Volvo Cars)

This is **PR1** of the alternative-names series. No API/search/UI behavior change yet beyond the schema field.

## Test plan
- [ ] Review migration SQL (`alternativeNames TEXT[] DEFAULT ARRAY[]::TEXT[]`)
- [ ] `npx jest src/lib/companyAlternativeNames.test.ts src/lib/companyLinkResolve.test.ts`
- [ ] Confirm match keys: `Volvo AB` / `VOLVO A.B.` / `AB Volvo` → `volvo`; `Volvo Cars` → `volvo cars`
- [ ] Confirm merge drops canonical duplicates and keeps distinct significant-token names

## Follow-ups
- API repo: mirror schema + company search/update
- Garbo: expose in API + resolve/search
- Garbo: auto-collect on confirmed link
- Validate: admin edit UI
- API: coverage/crawler search

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/lib/companyAlternativeNames.test.ts
+++ b/src/lib/companyAlternativeNames.test.ts
@@ -1,0 +1,83 @@
+import { describe, expect, it } from '@jest/globals'
+import {
+  isFormattingVariantOfCompanyName,
+  mergeAlternativeNames,
+} from './companyAlternativeNames'
+import { companyNameMatchKey } from './companyLinkResolve'
+
+describe('companyNameMatchKey', () => {
+  it('folds case, spaces, periods, and legal suffixes', () => {
+    expect(companyNameMatchKey('Volvo AB')).toBe('volvo')
+    expect(companyNameMatchKey('VOLVO  A.B.')).toBe('volvo')
+    expect(companyNameMatchKey('AB Volvo')).toBe('volvo')
+    expect(companyNameMatchKey('  Volvo,   AB ')).toBe('volvo')
+  })
+
+  it('keeps significant tokens so brand-family entities stay distinct', () => {
+    expect(companyNameMatchKey('Volvo AB')).toBe('volvo')
+    expect(companyNameMatchKey('Volvo Cars')).toBe('volvo cars')
+    expect(companyNameMatchKey('Volvo AB')).not.toBe(
+      companyNameMatchKey('Volvo Cars')
+    )
+  })
+
+  it('folds diacritics and ampersands', () => {
+    expect(companyNameMatchKey('Nestlé Sverige')).toBe('nestle sverige')
+    expect(companyNameMatchKey('H&M')).toBe('h m')
+    expect(companyNameMatchKey('H & M')).toBe('h m')
+  })
+})
+
+describe('mergeAlternativeNames', () => {
+  it('dedupes formatting variants and drops the canonical name', () => {
+    expect(
+      mergeAlternativeNames({
+        canonicalName: 'Volvo AB',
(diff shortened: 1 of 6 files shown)


src/lib/companyAlternativeNames.test.ts
```diff
@@ -0,0 +1,83 @@
...
+} from './companyAlternativeNames'
+import { companyNameMatchKey } from './companyLinkResolve'
+
+describe('companyNameMatchKey', () => {
+  it('folds case, spaces, periods, and legal suffixes', () => {
+    expect(companyNameMatchKey('Volvo AB')).toBe('volvo')
+    expect(companyNameMatchKey('VOLVO  A.B.')).toBe('volvo')
+    expect(companyNameMatchKey('AB Volvo')).toBe('volvo')
+    expect(companyNameMatchKey('  Volvo,   AB ')).toBe('volvo')
+  })
```
CarlWiren90, line 14: How do we think companies named "GE Paypal" would be handled for example? Where the "GE" here is a part of the name? We wouldn't want that to be just "Paypal"

CarlWiren90: I've had issued with companies named Group as well before. Just a thought here not really in regards to any of the above logic.

src/lib/companyAlternativeNames.test.ts
```diff
@@ -0,0 +1,83 @@
...
+    expect(companyNameMatchKey('Volvo AB')).toBe('volvo')
+    expect(companyNameMatchKey('Volvo Cars')).toBe('volvo cars')
+    expect(companyNameMatchKey('Volvo AB')).not.toBe(
+      companyNameMatchKey('Volvo Cars')
+    )
+  })
+
+  it('folds diacritics and ampersands', () => {
+    expect(companyNameMatchKey('Nestlé Sverige')).toBe('nestle sverige')
+    expect(companyNameMatchKey('H&M')).toBe('h m')
```
CarlWiren90, line 26: Do we actually want an H&M to be h m here?

CarlWiren90 (commented):
Just a general though. Our plan here is only to populate the alternativeNames array, and not using any of this to populate the actual companyName?

src/lib/companyAlternativeNames.test.ts
kaylawoodbury: hmmm 🤔  I see where you're coming from, but just to clarify, it's not that the name or alt name would be h m, but rather how the compare logic would adjust the formatting in order to do a comparison to see if there is a match

CarlWiren90: Yeah that is true. I think I'm still having some of the fuzzy matching in mind and the issues that can come from that. But with this fix we'd hopefully have a set to do an exact comparison against, and result to fuzzy if the first set of names doesn't match. Am I understanding correctly here?

CarlWiren90: So the H&M canonical would have an altNames array being like [HM, H M, H AND M] for example

PR #1418, opened 2026-09-17
Reuse Report.markdown when Chroma misses

kaylawoodbury:
## Summary
- On a Chroma miss, `parsePdf` now looks up stored `Report.markdown` (by `url` / `sourceUrl` / `pdfCache.publicUrl`) and re-indexes into Chroma without Docling.
- Falls back to the existing Docling → index → precheck flow when no markdown is in the registry.
- `forceReindex` still deletes the vector index and forces a full PDF re-parse; `callbackUrl` climate-plans path is unchanged.

## Test plan
- [ ] Clear stage Chroma (or delete one report’s Chroma entries) for a URL that already has `Report.markdown`
- [ ] Re-run that report without `forceReindex` and confirm logs show `markdown cache hit (postgres)` and Docling is skipped
- [ ] Confirm `indexMarkdown` + `precheck` still complete and extraction continues
- [ ] Re-run a URL with no `Report.markdown` and confirm Docling still runs (`markdown cache miss`)
- [ ] Confirm `forceReindex: true` still goes through Docling

Made with [Cursor](https://cursor.com)


eli6:
it looks fine! the naming of forceReindex is no longer perfect (it never was!) since it can refer to both reindexing in chroma and rerunning the full docling parse but that's ok for now i think

